PR #2, opened 2017-08-25
Updates sample deployment yml

caseyfw:
Brings OpenShift Shepherd deployment template in line with current s2i builder, provides a parameter to set an external DB and customise the UID/GID the web container should run as.


shepherd-openshift.yml
```diff
@@ -319,3 +307,11 @@ parameters:
   description: The admin users email address.
   required: true
   value: shepherd@shepherd.com
+- name: UID
```
singularo: Nice one.

PR #3, opened 2017-08-25
Feature/webops 231 streamline add site form

bart88:
Adds the following features : 

- Add local action for shp_site form
- Add javascript lib to create a custom event `inputdelay` to trigger after 600ms whilst a user is typing into a field. Drupal does not supply an event of this type.
- Avoid logic in the hook_form_alter by passing to a service
- Add ajax callback to generate a short name and ensure its uniqueness
- Attach callback to the site title field. Does not fire if you decide to manually edit. Node form submission validation will handle that. 

## How to test 
- `robo build && robo dev:content-generate` 
This will create a test site with the short name "test".
- Create an environment with the title : "test" 
- The short_name will be appened


singularo:
Testing in firefox & chrome:
* focus is stolen from the title and given to the short name, this shouldn't happen (its very annoying ;-)
* Duplicate protection only works for the second site with the same name, 3rd site with the same name still gets '-2' and wont save.
* Once you start editing the short name, should it then not update it again unless you blank it out, maybe.. Nitpicking with this one.

bart88:
I have updated this PR.

I have simplified the ajax and javascript down to just one ajax callback defined in the site service. I discovered an issue with ajax validation - needed an event that triggered after a reasonable time, for validating a users input whilst they are still in `:focus` on the field. 

This lead to the creation of a javascript lib that creates a custom event called `inputdelay` and triggers on all input text elements after to 600ms. 

Removed old javascript code that is not being used from the libraries.yml and the javascript directory.

singularo:
* Doesn't the short name need to be [0-9a-zA-Z\-] ?
* Focus stuff much nicer now.

bart88:
@singularo I opted for the short name to all be lowercase `[a-z0-9-]`

bart88:
Added form validation on the callback. Errors will be displayed in the `#prefix` element provided. User will need to manually correct the field if the short_name doesn't validate ( long site name ). Submission will then handle additional validation.

PR #9, opened 2017-09-01
Add redis support to shepherd.

singularo:
Add the setup for redis to shepherd for local dev


bart88:
It should be noted that you need to remove the `web/sites/default/settings.php`

PR #10, opened 2017-09-01
Ui improvements site backups

bart88:
Make the site environment operations ( restore and backup ) display in a modal window. 

Currently these small forms direct to a separate from then redirect back to the site node page which is a little disjointing. 

The user should be able to quickly trigger these operations whilst still keeping the context of the view they are on.


bart88:
Note the backup and restore is awaiting the queuing feature to work properly.

PR #12, opened 2017-09-04
Removed DATABASE_PASSWORD env var

caseyfw:
Fixes the relevant todo. Our shepherd-drupal-scaffold supports the `DATABASE_PASSWORD_FILE` env var so this one is no longer needed.


bart88:
:shipit:

PR #13, opened 2017-09-04
Webops-241-create-route-node-entity

bart88:
- Created a route entity.
- Created taxonomy that describes the type of route ( as defined in the json schema )
- Created view to display all routes 
- Created rest endpoint 
- Created local action to node add form and bound view to menu.


bart88:
Agreed. We need a short description for a route.

web/modules/custom/shp_routes/config/install/core.entity_form_display.node.shp_route.default.yml
```diff
@@ -0,0 +1,60 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.node.shp_route.field_shp_destination
+    - field.field.node.shp_route.field_shp_route_environment
```
caseyfw: Can you reuse the `field_shp_environment` field?

PR #15, opened 2017-09-05
Feature/webops 242 routes ui

bart88:
Create a UI that displays all the routes for a given site ( via related environment ).


web/modules/custom/shp_routes/shp_routes.links.action.yml
```diff
@@ -3,5 +3,14 @@ shp_routes.add_route:
   route_parameters:
     node_type: 'shp_route'
   title: 'Add new route'
+  options:
+    attributes:
+      class:
```
singularo: Sweet find.

PR #16, opened 2017-09-05
Added redis deployment config to template yaml

caseyfw:


shepherd-openshift.yml
```diff
@@ -10,18 +10,20 @@ metadata:
     openshift.io/display-name: Shepherd
     description: Provide a template that deploys an instance of the shepherd distribution.
     tags: php,drupal
-    template.openshift.io/documentation-url: https://gitlab.adelaide.edu.au/web-team/shepherd
+    template.openshift.io/documentation-url: 'https://gitlab.adelaide.edu.au/web-team/shepherd'
```
caseyfw: So apparently in yaml any value with a colon in it should be wrapped in single quotes.

shepherd-openshift.yml
```diff
@@ -44,14 +46,35 @@ objects:
...
+    name: shepherd-redis-is
+  spec:
+    lookupPolicy:
+      local: false
+    tags:
+      - annotations:
+          openshift.io/imported-from: 'redis:alpine'
+        from:
+          kind: DockerImage
+          name: 'redis:alpine'
```
caseyfw: @singularo in your shp_redis_support module this line is more like `name: docker.io/redis:alpine`. After some testing, the above is what openshift does when you just put a regular image ref into the "add to project" box.

shepherd-openshift.yml
singularo: Ahah.. Its a trap.

shepherd-openshift.yml
```diff
@@ -144,7 +169,7 @@ objects:
             value: /code/web/sites/default/files
           - name: SHEPHERD_INSTALL_PROFILE
             value: ${SHEPHERD_INSTALL_PROFILE}
-          image: ' '
+          image: 'shepherd-web-is:latest'
```
caseyfw: I'm not even sure why the image ref of `' '` worked originally.

shepherd-openshift.yml
```diff
@@ -186,22 +211,24 @@ objects:
       port: 3306
       targetPort: 3306
     selector:
-      name: shepherd-db
+      name: shepherd-db-dc
```
caseyfw: Now with more Smurfing :tm:

shepherd-openshift.yml
singularo: Err @caseyfw this content should point to this repo now. ;-)

shepherd-openshift.yml
```diff
@@ -211,8 +238,8 @@ objects:
             value: ${DATABASE_PASSWORD}
           - name: MYSQL_USER
             value: ${DATABASE_USER}
-          image: ''
-          name: mysql
+          image: 'openshift/mariadb:latest'
+          name: shepherd-db
```
caseyfw: This part looks really weird in contrast to the others in relation to the ImageChange trigger spec below. Here the image is specified as `openshift/mariadb:latest` but below it's called `mariadb:latest` and the namespace is in a separate key. Weird.

shepherd-openshift.yml
singularo: Ahah. Will fix there in another MR, thanks for that spotto.

PR #17, opened 2017-09-06
Specifying local lookup policy FALSE is actually what fixed this.

singularo:
Remove the no longer required docker.io


singularo:
Err, removing this

PR #18, opened 2017-09-06
Distribution view fix

caseyfw:
* Adds local action for add distribution to replace hacky header version.
* Adds missing fields to distro view.
* Changes path for add routes to be prettier.


bart88:
Looks good to me, builds and works 👍  :shipit:

PR #19, opened 2017-09-07
Feature/fix site environment view

bart88:
Fixes issues with with polling on the `site/{id}/environments` page
- Added README on how to develop with this module
- Update package json
- Add jsconfig ( for ides and editors that can read type definitions. see visual studio code )
- Refactor the Table.vue component into more logical functions, fix the interval issues with reconnecting and catch ajax request errors ( also disconnect if hits error limit )


web/modules/custom/shp_vue_view/README.md
```diff
@@ -0,0 +1,13 @@
+# Shepherd Vue.js View Module
```
pingers: This doc should be called DEVELOPERS.md

web/modules/custom/shp_vue_view/js/src/table/Table.vue
```diff
@@ -151,6 +183,33 @@
...
+        },
+        /**
+         * Disconnect the polling interval.
+         *
+         * @param {string} msg - Log message giving reason for disconnection.
+         */
+        disconnectPollingTimer(msg) {
+          clearInterval(poller);
+          poller = null;
+          console.info('Disconnecting polling : ' + msg);
```
pingers: I think the console logging is fine for now - but maybe a future thing would only enable console logging when some debug mode is enabled.

pingers:
Well, I tried it....

I was on the environments page for a site... status went from "Unknown" to "Building". Cool.
I did the site install, but the page never updated to show the new state "Running". We probably give up trying to find the state too quickly - especially when building, which should fairly quickly pass/fail.

web/modules/custom/shp_vue_view/js/src/table/Table.vue
```diff
@@ -151,6 +183,33 @@
...
+          if (reconnectPollerTimer) {
+            clearTimeout(reconnectPollerTimer);
+            reconnectPollerTimer = null;
+          }
+          reconnectPollerTimer = setTimeout(() => {
+            console.info('Restarting polling');
+            pollTiming = POLL_UPDATE_CYCLE;
+            disconnect_count = 0;
+            poller = setInterval(() => { context.poll(); }, pollTiming);
+          }, 900000);
```
pingers: The retries are const's, but not the poll interval?

bart88: The interval is a const but the reconnection timeout wasn't. Changed this.

bart88:
Updated the unchanged state timeout. This should wait a little while longer during a build before disconnecting for 15 minutes.

pingers:
Yeah, this should help. Cheers! :ship:

PR #21, opened 2017-09-08
Removed ldap madness

caseyfw:
Makes this project actually usable outside the uni staff subnet.


singularo:
Yeah, lets get this stuff out of the public repo, it stops it being usable OOTB.
We might need some locally, but we'll handle that then.

PR #23, opened 2017-09-08
Feature/refactor redis to environment

singularo:
The shp_redis_support module was incorrectly calling the openshift client directly, re-factor to use the client via the service, which should support other providers.

Also add in the support required for setting environment variables so that redis deployments work correctly.


singularo:
So this now work, but brings up as many questions as it resolves @pingers & @bart88 :
* Since there is now another option, its possible to change orchestration providers between deployments, meaning you can get mixed deployments.
* If you switch orchestration providers and try and remove a deployment, things go bad as shepherd just uses the 'current' orchestration provider.
* Should we lock to a single type so that this is not an issue, or;
* Add something so each environment needs to know what type of orchestration provider it was deployed by.
* Should we store the deployment type as part of the site and lock that site to a particular type? or as the environment, which will make things very dynamic.

pingers:
Yeah, making significant provider changes with active deployments will be painful for now.

We should look at using decorator pattern to add functionality like redis to the orchestration provider. https://symfony.com/doc/current/service_container/service_decoration.html
https://en.wikipedia.org/wiki/Decorator_pattern

...and where the orchestration provider settings attach to becomes the next issue :smile: I.e. Site? Distro? Env? Aargh.

web/modules/custom/shp_redis_support/src/Plugin/OrchestrationProvider/OpenShiftWithRedis.php
```diff
@@ -0,0 +1,155 @@
...
+            ],
+            'type' => 'ImageChange',
+          ],
+        ],
+      ],
+    ];
+
+    $this->client->createDeploymentConfig($redis_deployment_config);
+    $this->client->createService($redis_name, $redis_name, $redis_port, $redis_port);
+
```
pingers: Extra space

web/modules/custom/shp_orchestration/src/Event/OrchestrationEnvironmentEvent.php
```diff
@@ -7,50 +7,90 @@
...
+  /**
+   * Storage to pass env vars around
+   */
+  protected $environmentVariables;
 
   /**
    * Constructs a Orchestration deployment event object.
    *
-   * @param @var \UniversityOfAdelaide\OpenShift\Client $client
+   * @param $orchestrationProvider
```
bart88: These params should have declared types.

web/modules/custom/shp_orchestration/src/Service/Environment.php
```diff
@@ -90,7 +92,15 @@ public function created(NodeInterface $node) {
     $env_vars = $this->configuration->getEnvironmentVariables($node);
     $secrets = $this->configuration->getSecrets($node);
 
-    return $this->orchestrationProviderPlugin->createdEnvironment(
+    // Allow other modules to react to the Environment creation.
+    $eventDispatcher = \Drupal::service('event_dispatcher');
+    $event = new OrchestrationEnvironmentEvent($this->orchestrationProviderPlugin, $deployment_name, NULL);
```
bart88: Why are you passing NULL as the third argument ?

web/modules/custom/shp_orchestration/src/Service/Environment.php
```diff
@@ -143,11 +159,24 @@ public function deleted(NodeInterface $node) {
       return FALSE;
     }
 
-    return $this->orchestrationProviderPlugin->deletedEnvironment(
+    $deployment_name = $this->orchestrationProviderPlugin::generateDeploymentName(
```
bart88: Why are you calling this method statically ?

web/modules/custom/shp_orchestration/src/Service/Environment.php
```diff
@@ -143,11 +159,24 @@ public function deleted(NodeInterface $node) {
...
+    );
+
+    $result = $this->orchestrationProviderPlugin->deletedEnvironment(
       $distribution->title->value,
       $site->field_shp_short_name->value,
       $node->id()
     );
+
+    // Allow other modules to react to the Environment deletion.
+    $eventDispatcher = \Drupal::service('event_dispatcher');
```
bart88: 💉  Dependancy Injection

web/modules/custom/shp_redis_support/src/EventSubscriber/DeploymentEventSubscriber.php
```diff
@@ -12,140 +12,32 @@ class DeploymentEventSubscriber implements EventSubscriberInterface {
    * {@inheritdoc}
    */
   public static function getSubscribedEvents() {
+    $events[OrchestrationEvents::SETUP_ENVIRONMENT][]   = array('setupRedisDeployment');
```
bart88: Excess space

web/modules/custom/shp_redis_support/src/EventSubscriber/DeploymentEventSubscriber.php
```diff
@@ -12,140 +12,32 @@ class DeploymentEventSubscriber implements EventSubscriberInterface {
...
     $events[OrchestrationEvents::CREATED_ENVIRONMENT][] = array('createRedisDeployment');
     $events[OrchestrationEvents::DELETED_ENVIRONMENT][] = array('deleteRedisDeployment');
 
     return $events;
   }
 
+  public function setupRedisDeployment(OrchestrationEnvironmentEvent $event) {
```
bart88: Docblock

web/modules/custom/shp_redis_support/src/EventSubscriber/DeploymentEventSubscriber.php
```diff
@@ -155,19 +47,11 @@ public function createRedisDeployment(OrchestrationEnvironmentEvent $event) {
...
-
-    //$client->updateDeploymentConfig($redis_name, 0);
-    //$client->updateReplicationControllers('', 'app=' . $redis_name, 0);
-
-    $client->deleteService($redis_name);
-
-    $client->deleteDeploymentConfig($redis_name);
-    //$client->deleteReplicationControllers('', 'app=' . $redis_name);
+    $orchestration_provider = $event->getOrchestrationProvider();
+    if ($orchestration_provider->getPluginId() == 'openshift_with_redis') {
```
bart88: Strict type check

web/modules/custom/shp_orchestration/src/Service/Environment.php
pingers: :man_dancing: :dancer: :dancing_men: :dancing_women:

pingers:
:ship: We've deferred refactoring for now. Jira WEBOPS-250 is the issue for that.

Issue #22, opened 2017-09-08
Default project to be built in ShepherdContentGenerate.php is not externally accessible


caseyfw:

singularo:
Indeed.

pingers:
Probably need an env var to override for convenient local dev?

singularo:
Done. Still more UA specific things to remove, but getting there.

PR #24, opened 2017-09-10
Feature/use public repository for default

singularo:
Work on issue #22 
Use public repo as the default generated site
* Remove the ua specific default distribution config
* Fix NULL/string issue with updatedDistribution function
* Add overridable env var to specify the repository
* Add support for env var to content generate script


web/modules/custom/shp_content_types/config/install/field.field.node.shp_distribution.field_shp_git_repository.yml
```diff
@@ -14,7 +14,7 @@ required: true
 translatable: false
 default_value:
   -
-    value: 'git@gitlab.adelaide.edu.au:web-team/ua-wcms-d8.git'
+    value: 'https://github.com/universityofadelaide/shepherd-example-drupal.git'
```
pingers: This contents of this repo should have been creating using the project command...
`composer create-project universityofadelaide/shepherd-drupal-project shepherd-example-drupal --stability dev --no-interaction`

pingers: ...and this makes me think we really should rename `distribution` to `project`. Would make a lot more sense outside of Drupal. Totally unrelated to this PR 😄, but thought I better write it down somewhere.

singularo: Yep, they were, everything else gets written by the composer update/install, I guess it comes down to whether we want to commit those things that would get overwritten.

web/modules/custom/shp_orchestration/src/OrchestrationProviderBase.php
```diff
@@ -83,4 +83,17 @@ public function getConfigEntity() {
...
+   * Converts a string into a format acceptable for orchestration providers.
+   *
+   * @param string $text
+   *   The title to be sanitised.
+   *
+   * @return string
+   *   sanitised title.
+   */
+  static function sanitise($text) {
+    return strtolower(preg_replace('/[\/\s]+/', '-', $text));
```
pingers: Do we need another machine name converter/generator? *grumble grumble*

pingers: We used transliteration elsewhere for this.

singularo: I agree, but that uses preg_replace as well. Hmm maybe we should make a service, this just moved the existing function, lets make fixing that a seperate PR.

pingers: I looked again for a core issue about it... no such luck.

ShepherdContentGenerate.php
```diff
@@ -14,6 +14,9 @@
 $domain_name = getenv("OPENSHIFT_DOMAIN") ?: '192.168.99.100.nip.io';
 $openshift_url = getenv("OPENSHIFT_URL") ?: 'https://192.168.99.100:8443';
 $token = trim(getenv("TOKEN"));
+
+$example_repository = getenv("EXAMPLE_REPOSITORY");
```
pingers: If this is mandatory, then we should die and show a message ASAP.

pingers: Or just set a default and remove it from dsh.project.

singularo: Adjusted so that's the case. Leaving in dsh.project so it gets passed into the shell, so that the script can detect it.

pingers:
:ship:

PR #25, opened 2017-09-12
Feature/remove dud config

pingers:
Speaks for itself :)

* Remove bartik we don't want/need.
* Move UA specific config into the UA install profile - for later removal.
* Small copy/pasta typo fixes.


singularo:
Nice one, Merge away

PR #26, opened 2017-09-13
Feature/fix annotation grouping

singularo:
* Add new parameter for the parent application to the create service call
* Sneak in some validation so the shp_redis_support supports removing other orchestration provider created applications.
Requires openshift-client PR to come any minute...


pingers:
:ship:

PR #27, opened 2017-09-15
Feature/webops 244 provision users from ldap

singularo:
* Remove the last UA specific parts of the repo in preparation for adding the UA only parts in the private repo.
* Adjust directories for renaming the modules into one shepherd module 'set'
* Add script that can split and update the shepherd-modules repo - https://github.com/universityofadelaide/shepherd-modules using https://github.com/splitsh/lite
* Add composer.json for the shepherd module 'set'.

Once this is merged, we'll be able to pull in the module 'set' for our local shepherd version and add the custom ldap features alongside.


pingers:
I take it the web/modules/custom/shepherd is the for subtree split?

singularo:
Yes, it can work without that, but then the composer.json for shepherd would be in modules/custom, which is weird. Those bits are currently missing from the shepherd-modules repo due to that.

singularo:
Note that the script will not work until this MR is in as the web/modules/custom/shepherd dir wont be there yet.

singularo:
Actually, if I use the branch name, it will.

singularo:
Going to get rid of this and do it differently.

PR #28, opened 2017-09-18
Feature/webops 245 rename distribution to project

pingers:
"Project" reflects the nature of this configuration better than distribution.

Renaming some of the config was fun :-1:


singularo:
Seems good to me :shipit:

PR #31, opened 2017-09-20
Fix recent install breakage.

singularo:
The tags/taxonomy things you changed actually stopped things from installing correctly @pingers, this fixes that issue.


bart88:
Builds and works 👍 🚢

pingers:
*shrugs* Not sure how moving the views fixes it. Will chat Monday :)

PR #32, opened 2017-09-27
PHPCS clean up.

pingers:
Just comments and phpcs clean up.


web/modules/custom/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
```diff
@@ -144,6 +144,8 @@ public function createdEnvironment(
     array $secrets = [],
     array $cron_jobs = []
   ) {
+    // @todo Refactor this. _The complexity is too damn high!_
```
singularo: Noice

PR #38, opened 2017-10-09
Feature/dynamic service routing

singularo:
Introduce basic A/B style deployments for shepherd.

To test:
1. Create a new site with a 'prod' url. Shepherd will now create a service and route for that site - check in openshift.
2. Create a some new dev environments for that site.
3. Check in openshift, the original site service and route will be there as well as new ones for the dev environment. The dev one will work, the production one wont yet work.
4. On the environments page, from the drop-down, promote one of the environments
5. Both endpoints will now be associated with the environment selected.
6. Check in openshift, you can see the services have been updated.
7. Promote the other environment, it will switch back.

Also implemented in this MR
url on the site page
taxonomy changes to support the promotion/demotion
taxonomy changes to stop string comparing and use fields


web/modules/custom/shepherd/shp_content_types/config/install/field.field.taxonomy_term.shp_environment_types.field_shp_protect.yml
```diff
@@ -0,0 +1,22 @@
+langcode: en
```
pingers: Can't we just do this through permissions? I.e. we're moving to archiving.
Or the purpose is more around preventing an environment from being archived? spinning down pods, removing routing etc. If so, I'm a little confused, because this functionality was built into the Production tag I believe. So now, there would be 2 ways you can prevent an environment from being archived. Do we combine the two and make the "protect" field disabled while the environment tag is "Production"?

web/modules/custom/shepherd/shp_content_types/config/install/field.field.taxonomy_term.shp_environment_types.field_shp_update_go_live.yml
```diff
@@ -0,0 +1,22 @@
...
+dependencies:
+  config:
+    - field.storage.taxonomy_term.field_shp_update_go_live
+    - taxonomy.vocabulary.shp_environment_types
+id: taxonomy_term.shp_environment_types.field_shp_update_go_live
+field_name: field_shp_update_go_live
+entity_type: taxonomy_term
+bundle: shp_environment_types
+label: 'Update go live'
+description: ''
```
pingers: This needs a description.

"Set the initial go-live date for a site, when an environment receives this tag, if it has not already been set."

web/modules/custom/shepherd/shp_custom/shp_custom.module
```diff
@@ -223,12 +213,13 @@ function shp_custom_node_delete(EntityInterface $entity) {
...
  */
-function shp_custom_node_access(EntityInterface $entity, $operation, AccountInterface $account) {
-  if ($entity->bundle() == 'shp_environment') {
-    if ($operation == 'delete') {
-      // @todo Use field_shp_environment_type term association to determine access.
-      if ($entity->field_shp_machine_name->value === 'prd') {
-        return AccessResult::forbidden();
+function shp_custom_node_access(NodeInterface $node, $op, AccountInterface $account) {
+  if ($node->bundle() == 'shp_environment') {
+    if ($op == 'delete') {
```
pingers: === both ifs while you're at it :+1:

web/modules/custom/shepherd/shp_custom/shp_custom.module
```diff
@@ -223,12 +213,13 @@ function shp_custom_node_delete(EntityInterface $entity) {
...
+      if ($environment_term = Term::load($node->field_shp_environment_type->target_id)) {
+        if ($environment_term->field_shp_protect->value) {
+          return AccessResult::forbidden();
+        }
       }
     }
```
pingers: If we have no opinion, we should be doing ```return AccessResult::neutral();```

web/modules/custom/shepherd/shp_custom/shp_custom.routing.yml
```diff
@@ -53,3 +53,21 @@ shp_custom.admin_config_shepherd:
...
   requirements:
     _permission: 'access administration pages'
+shp_custom.environment-promote-form:
+  path: '/site/{site}/environment/{environment}/promote'
+  defaults:
+    _form: '\Drupal\shp_custom\Form\EnvironmentPromoteForm'
+    _title_callback: '\Drupal\shp_custom\Form\EnvironmentPromoteForm::getPageTitle'
+    _title: 'Promote'
+  options:
+    _auth: [ 'basic_auth', 'cookie' ]
```
pingers: Why different from the defaults?

web/modules/custom/shepherd/shp_custom/src/Form/EnvironmentPromoteForm.php
```diff
@@ -0,0 +1,151 @@
...
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+use Drupal\node\NodeInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * EnvironmentPromoteForm.
+ *
+ * Allows promoting environments.
+ *
+ * @package Drupal\shp_custom\Form
```
pingers: Not needed.

web/modules/custom/shepherd/shp_custom/src/Form/EnvironmentPromoteForm.php
```diff
@@ -0,0 +1,151 @@
...
+      '#type' => 'item',
+      '#title' => $this->t('@production_service', ['@production_service' => $site->field_shp_domain->value]),
+      '#description' => $this->t('Will now be handled by'),
+    ];
+    $form['environment_service'] = [
+      '#type' => 'item',
+      '#title' => $this->t('@environment_service', ['@environment_service' => $environment->getTitle()]),
+    ];
+
+    // @todo everything is exclusive for now, implement non-exclusive?
```
pingers: How could we send traffic to multiple environments at the same domain+path? I'm missing the use case :smile:

web/modules/custom/shepherd/shp_custom/src/Form/EnvironmentPromoteForm.php
```diff
@@ -0,0 +1,151 @@
...
+    $site = $form_state->get('site');
+    $environment = $form_state->get('environment');
+    $exclusive = TRUE; // $form_state->getValue('exclusive');
+
+    if ($result = \Drupal::service('shp_orchestration.environment')->promoted($site, $environment, $exclusive)) {
+      drupal_set_message($this->t('Promoted %environment for %site successfully', [
+        '%environment' => $environment->getTitle(),
+        '%site' => $site->getTitle(),
+      ]));
+      \Drupal::service('shp_custom.site')->checkGoLiveApplied($environment);
```
pingers: Inject the service.

web/modules/custom/shepherd/shp_custom/src/Service/Environment.php
```diff
@@ -179,4 +180,39 @@ protected function loadTaxonomyTerm($tid) {
...
+      $operations['promote'] = [
+        'title'      => $this->t('Promote'),
+        'weight'     => 1,
+        'url'        => Url::fromRoute('shp_custom.environment-promote-form',
+          ['site' => $site, 'environment' => $environment]),
+        // Render form in a modal window.
+        'attributes' => [
+          'class'               => ['button', 'use-ajax'],
+          'data-dialog-type'    => 'modal',
+          'data-dialog-options' => Json::encode(['width'  => '50%',
```
pingers: Keys and values should be on a new line.

web/modules/custom/shepherd/shp_custom/src/Service/Environment.php
```diff
@@ -179,4 +180,39 @@ protected function loadTaxonomyTerm($tid) {
...
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   Entity to apply operations to.
+   */
+  public function operationsLinksAlter(array &$operations, EntityInterface $entity) {
+    $site = $entity->field_shp_site->target_id;
+    $environment = $entity->id();
+    $environment_term = Term::load($entity->field_shp_environment_type->target_id);
+
+    // If its not a protected environment, it can be promoted.
+    if (!$environment_term->field_shp_protect->value) {
```
pingers: Protected means it can't be changed at all? I'm confused.

web/modules/custom/shepherd/shp_custom/src/Service/Site.php
```diff
@@ -65,9 +65,9 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager) {
    *   TRUE if applied go live date.
    */
   public function checkGoLiveApplied(Node $environment) {
```
pingers: This method name is terrible. It applies the go live date, not just checking if it was applied.

applyGoLiveDate() ?

web/modules/custom/shepherd/shp_custom/src/Form/EnvironmentPromoteForm.php
pingers: It's reaaaaally fricken weird that the form is responsible for applying the go live date to the site.
This should be triggered as part of entity API.

web/modules/custom/shepherd/shp_orchestration/src/OrchestrationProviderTrait.php
```diff
@@ -0,0 +1,26 @@
...
+use Drupal\node\NodeInterface;
+
+trait OrchestrationProviderTrait {
+
+  /**
+   * @param \Drupal\node\NodeInterface $site
+   *
+   * @return \Drupal\node\NodeInterface|bool
+   */
+  protected function getProjectFromSite(NodeInterface $site) {
```
pingers: This method belongs on the Site service.

pingers: ..and getSiteFromEnvironment() is on the wrong Environment service. Grrr.

They belong on the Drupal\shp_custom\Service namespaced services.

web/modules/custom/shepherd/shp_orchestration/src/Service/Environment.php
```diff
@@ -192,4 +174,76 @@ public function deleted(NodeInterface $node) {
...
+
+    return $result;
+  }
+
+  /**
+   * @param \Drupal\node\NodeInterface $environment
+   *
+   * @return \Drupal\node\NodeInterface|bool
+   */
+  protected function getSiteFromEnvironment(NodeInterface $environment) {
```
pingers: Pretty sure this isn't a orchestration related method, so it probably belongs on the shp_custom Environment service.

web/modules/custom/shepherd/shp_content_types/config/install/field.field.taxonomy_term.shp_environment_types.field_shp_protect.yml
singularo: It was string matching on the tag name. What if people want to call it Live instead of Production, just using a field, the field is checked elsewhere in this MR where before it was a string comparison on the tag name. Using the field completely replaces the string matching.

web/modules/custom/shepherd/shp_custom/shp_custom.module
singularo: Done

web/modules/custom/shepherd/shp_custom/shp_custom.module
singularo: Err, it already does that , last line in function..

web/modules/custom/shepherd/shp_custom/shp_custom.routing.yml
singularo: Just a copy/paste from the way the backup dropdowns are done, all should be changed if its no right.

web/modules/custom/shepherd/shp_custom/src/Form/EnvironmentPromoteForm.php
singularo: Maybe, lots of it in the rest of the source though..  Do we want to remove them all?

web/modules/custom/shepherd/shp_custom/src/Form/EnvironmentPromoteForm.php
singularo: A/B is where you typically "add" to a live site, so you could promote an env but be not exclusive, and then 50% (or whatever, depending on settings) would be using old vs new code.

As we might have db changes, possibly never an issue for us, but wouldn't be hard to add in.

https://docs.openshift.com/container-platform/3.3/dev_guide/deployments/advanced_deployment_strategies.html

pingers:
:ship: I think it's time!

There's some follow up issues... but not much to do with this PR.

PR #39, opened 2017-10-11
Removing the last UA specific parts.

singularo:
There is not much UA specific left, this removes the install profile, which is now in the UA gitlab fork.
There will need to be a MR there that adds in instructions on updating from this repo, which we can do separately.


RoboFile.php
```diff
@@ -28,7 +28,7 @@ public function __construct() {
    * {@inheritdoc}
    */
   protected function getDrupalProfile() {
```
pingers: Can't this go in favour of an env var?

pingers:
All looks fair to me :ship:

PR #41, opened 2017-10-17
Add the log_stdout module, for aggregating logs with containerisation.

singularo:


pingers:
:ship:  Don't forget to composer update.

PR #42, opened 2017-10-17
Change roles and names so that things work as expected.

singularo:


pingers:
Makes sense... are all the others correct and match the wcms-d8 distro names?

Guess we'll have to figure out how to move these roles elsewhere... and make the module that consumes the roles from shepherd public if it's not already.

pingers:
_I think you've done... very well!_ :shipit:

PR #45, opened 2017-10-17
Add enviornment site/path constraint.

singularo:
Add constraint that looks at the path and domain to the environment or its possible to create environments that are unreachable.


pingers:
I like. :shamrock:

PR #46, opened 2017-10-17
Show the url based on the env type in the environment view.

singularo:
Remove the call to the orch provider and use the built in vars we already have as well as detect env type and show appropriate url.

### DIFF
--- a/web/modules/custom/shepherd/shp_custom/src/Plugin/views/field/SiteEnvironmentUrl.php
+++ b/web/modules/custom/shepherd/shp_custom/src/Plugin/views/field/SiteEnvironmentUrl.php
@@ -2,6 +2,9 @@
 
 namespace Drupal\shp_custom\Plugin\views\field;
 
+use Drupal\Core\Link;
+use Drupal\Core\Url;
+use Drupal\taxonomy\Entity\Term;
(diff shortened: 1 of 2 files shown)


web/modules/custom/shepherd/shp_custom/src/Plugin/views/field/SiteEnvironmentUrl.php
```diff
@@ -26,36 +29,24 @@ public function query() {
...
-        '#type' => 'link',
-        '#title' => substr($url->toString(), 2),
-        '#url' => $url,
-      ];
+    $environment_term = Term::load($entity->field_shp_environment_type->target_id);
+
+    // If its a protected environment, its production, show that url.
+    if ($environment_term->field_shp_protect->value) {
+      $site = $entity->field_shp_site->first()->entity;
+      $domain_and_path = rtrim($site->field_shp_domain->value . $site->field_shp_path->value, '/');
```
pingers: I don't think we should rtrim. Meh. If people want/don't want slashes, don't put them in.

pingers:
I reckon this is good for now. Let's fix the notifications and put them back in later.

singularo:
All ready now.

PR #47, opened 2017-10-17
Feature/default project branch

pingers:
Fresh shepherd install required.

Run the content generate and try creating a project with a different default branch and check the add environment form uses that value.


web/modules/custom/shepherd/shp_custom/src/Service/Environment.php
```diff
@@ -99,16 +108,32 @@ public function formAlter(array &$form, FormStateInterface $form_state) {
...
+   *
+   * @param array $form
+   *   Form render array.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   Form state.
+   */
+  public function setBranchField(array &$form, FormStateInterface $form_state) {
+    if ($site = $form['field_shp_site']['widget'][0]['target_id']['#default_value']) {
+      $project = $this->site->getProject($site);
+      $form['field_shp_git_reference']['widget'][0]['value']['#default_value'] = $project->field_shp_git_default_ref->value;
```
fenetikm: instead of this type of madness, which I have in the past been guilty of, could one instead do something like:

```
$node = Node::create([
  ...
)];
$node->set('field_shp_git_reference', 'your lovely value');
$form = \Drupal::service('entity.form_builder')->getForm($node, 'default');
```
?

Setting via the form array seems so wrong town.

fenetikm: as you were.

web/modules/custom/shepherd/shp_content_types/config/install/core.entity_form_display.node.shp_project.default.yml
```diff
@@ -21,46 +22,31 @@ dependencies:
     - key_value_field
 third_party_settings:
   field_group:
-    group_shp_probes:
```
singularo: Did you mean to revert this stuff?

PR #49, opened 2017-10-18
Add tmp dir to docker-compose, fix incorrect public dir setting.

singularo:
Add tmp dir support and fix public_dir support so it means what its supposed to.
Basically, its the mount point, it will always appear at /code/web/sites/default/files
Other MR's related to come.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,8 @@ services:
       DATABASE_USER: user
       DATABASE_PASSWORD: password
       PRIVATE_DIR: /shared/private
-      PUBLIC_DIR: web/sites/default/files
+      TMP_DIR: /shared/tmp
+      PUBLIC_DIR: /shared/public
       HASH_SALT: random-hash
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -17,7 +17,8 @@ services:
       DATABASE_USER: user
       DATABASE_PASSWORD: password
       PRIVATE_DIR: /shared/private
-      PUBLIC_DIR: web/sites/default/files
+      TMP_DIR: /shared/tmp
+      PUBLIC_DIR: /shared/public
```
pingers: Wouldn't this still rely on a symlink that lives with the webroot to work? I.e. I thought public files had to live within the webroot to work.

singularo: Correct, so the PUBLIC_DIR allows you to specify where the back end stores things, but it will always be symlinked into web/sites/default/files on the Drupal pods.

Line 24 in files/apache2-foreground - https://github.com/universityofadelaide/docker-apache2-php7/pull/12/files

pingers:
:ship:

PR #50, opened 2017-10-19
Add small change and doc to install profile.

singularo:
Docs a bit rough, gave it a crack.

### DIFF
--- a/web/profiles/shepherd/shepherd.install
+++ b/web/profiles/shepherd/shepherd.install
@@ -56,9 +56,10 @@ function shepherd_install() {
    * Note that we have chosen to use the Environment variable simply as the
    * initial default here, rather than write it out in $config in settings.php.
    * https://www.drupal.org/node/2662646 has some useful information, but the
-   * main problem with overriding in settings.php is that the ui and config
-   * system then doesn't know what is going on, and you can try setting it there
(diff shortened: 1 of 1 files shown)


singularo:
We will need to do this in ua-wcms-d8:shepherd as well

web/profiles/shepherd/shepherd.install
```diff
@@ -52,6 +52,16 @@ function shepherd_install() {
   // Enable the admin theme.
   \Drupal::configFactory()->getEditable('node.settings')->set('use_admin_theme', TRUE)->save(TRUE);
 
+  /*
+   * Note that we have chosen to use the Environment variable simply as the
+   * initial default here, rather than write it out in $config in settings.php.
+   * https://www.drupal.org/node/2662646 has some useful information, but the
+   * main problem with overriding in settings.php is that the ui and config
```
pingers:
```
  * main problem with overriding in settings.php is that the config system
  * and file system API's become out of sync. Only the file system API 
  * understands the $settings.php override and using config API doesn't
  * return the value defined in settings.php config.

pingers:
Functionally shippable - just a little doc tweak. Cheers! :shipit: :rocket:

PR #52, opened 2017-10-19
Fix the site machine name field auto-fill.

pingers:
Uh, magic! Try the add site form.


singularo:
Awesome, much simpler ;-)
:shipit:

PR #55, opened 2017-10-26
Reinstate the cron jobs.

singularo:
Also needs - https://github.com/universityofadelaide/openshift-client/pull/16


web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
```diff
@@ -212,15 +212,14 @@ public function createdEnvironment(
...
-          //$this->client->createCronJob(
-          //  $deployment_name . '-' . \Drupal::service('shp_custom.random_string')->generate(5),
-          //  $image_stream['status']['dockerImageRepository'] . ':' . $source_ref,
-          //  $schedule,
-          //  $args_array,
-          //  $volumes,
-          //  $deploy_data
-          //);
+          $this->client->createCronJob(
+            $deployment_name . '-' . \Drupal::service('shp_custom.random_string')->generate(5),
```
bart88: Inject this.

web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
```diff
@@ -63,7 +63,7 @@ public function createdProject(string $name, string $builder_image, string $sour
     ];
 
     try {
-      $image_stream = $this->client->generateImageStreamConfig($sanitised_name);
+      $image_stream = $this->client->generateImageStream($sanitised_name);
```
pingers: Not sure this is correct. generateImageStreamConfig() is the right method.

see https://github.com/universityofadelaide/openshift-client/blob/develop/src/Client.php#L736

web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
singularo: another MR - how do i inject to a plugin?

singularo:
Ok, this should be ready now.

web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
pingers: You'll have to use `implements ContainerInjectionInterface` and implement the create() method.
Something like
```
  public static function create(ContainerInterface $container) {
    return new static($container->get('shp_custom.random_string'));
  }
```

fenetikm: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Menu%21LocalActionDefault.php/8.2.x

pingers:
How do we see that cron ran from openshift? Is the only place you can view cronjobs in the monitoring log?

pingers:
:shipit:

PR #56, opened 2017-10-26
Feature/add direct logs

bart88:
Implement a direct to logs link similar to the direct to terminal link on an environment.


singularo:
Doesn't work with a fresh install:
```
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The service        [error]
"shp_custom.environment" has a dependency on a non-existent service
"plugin.manager.orchestration_provider". in
[snip]
➜  APP_ERROR: drush site-install failed.
🐳 simon@shepherd:/code$ git status
On branch feature/add-direct-logs
```

singularo:
Should also look at disabling the caching in the view, as that makes this and the terminal link one 'feel' buggy, but all they need is a cache clear.

singularo:
Lets also add target=_blank so we get a new tab for the terminal and logs.

singularo:
But yeah, all works ok now, just a couple of suggestions.

web/modules/custom/shepherd/shp_custom/src/Service/Environment.php
```diff
@@ -86,14 +95,20 @@ class Environment {
...
+                              AccountProxyInterface $currentUser,
+                              Site $site) {
     $this->requestStack = $requestStack;
     $this->entityTypeManager = $entityTypeManager;
     $this->currentRequest = $this->requestStack->getCurrentRequest();
     $this->node = $this->entityTypeManager->getStorage('node');
     $this->taxonomyTerm = $this->entityTypeManager->getStorage('taxonomy_term');
     $this->currentUser = $currentUser;
     $this->site = $site;
+    // @todo - too many cross dependancies on this service, causing install failures. Fix.
```
pingers: Needs refactor :+1:

web/modules/custom/shepherd/shp_custom/src/Service/Environment.php
```diff
@@ -259,12 +275,24 @@ public function operationsLinksAlter(array &$operations, EntityInterface $entity
         'url'        => $terminal,
       ];
     }
+
+    if ($logs) {
+      $operations['logs'] = [
+        'title' => $this->t('Logs'),
+        'weight' => 4,
+        'url' => $logs,
```
pingers: Nitpick: indenting is different from the terminal operation above.

pingers:
It works. :ship:

PR #57, opened 2017-10-27
Update deps, fixes group issue.

bart88:
Updating the group module fixes the following issue: 
https://www.drupal.org/node/2908830

This was breaking any site, environment or project delete.


singularo:
Fixed the issue for me.

PR #62, opened 2017-11-01
Feature/remove paragraphs views

pingers:
Just some clean up of paragraphs module.


bart88:
Happy for this to 🚢  There is still more of this dead code floating around. Good start.

PR #63, opened 2017-11-03
Remove the orchestration provider from constructor

bart88:
The orchestration provider object has a closure or db connection in it that is causing the following : 

`Exception: Serialization of &amp;#039;Closure&amp;#039; is not allowed in serialize() (line 14 of /code/web/core/lib/Drupal/Component/Serialization/PhpSerialize.php)` 

This is due to the Environment service altering a form with an ajax callback that calls a method on the service itself. The formbuilder appears to storing this object ( Environment service ) in cache and serialising. Causing the above exception in the logs.


bart88:
Closing in favour of #66

PR #64, opened 2017-11-03
Static methods can't call non-static methods. Refactors unique short …

pingers:
…name check.

Creating a site caused php exception... this should fix it.

### DIFF
--- a/web/modules/custom/shepherd/shp_custom/shp_custom.module
+++ b/web/modules/custom/shepherd/shp_custom/shp_custom.module
@@ -299,3 +299,21 @@ function shp_custom_entity_operation_alter(array &$operations, EntityInterface $
     \Drupal::service('shp_custom.environment')->operationsLinksAlter($operations, $entity);
   }
 }
+
+/**
+ * Ensure that the short_name generated is unique.
+ *
+ * @param string $short_name
+ *   Generated short name.
+ *
+ * @return string
+ *   If not unique adds a number to end of string, otherwise valid.
+ */
+function shp_custom_generate_unique_short_name($short_name) {
+  $results = \Drupal::service('shp_custom.site')->loadEntitiesByFieldValue('shp_site', 'field_shp_short_name', $short_name);
+  if ($results) {
+    $count = count($results);
+    return $short_name . '-' . $count;
+  }
+  return $short_name;
+}
--- a/web/modules/custom/shepherd/shp_custom/shp_custom.module
+++ b/web/modules/custom/shepherd/shp_custom/shp_custom.module
@@ -310,10 +310,5 @@ function shp_custom_entity_operation_alter(array &$operations, EntityInterface $
  *   If not unique adds a number to end of string, otherwise valid.
  */
 function shp_custom_generate_unique_short_name($short_name) {
-  $results = \Drupal::service('shp_custom.site')->loadEntitiesByFieldValue('shp_site', 'field_shp_short_name', $short_name);
-  if ($results) {
-    $count = count($results);
-    return $short_name . '-' . $count;
-  }
-  return $short_name;
+  return \Drupal::service('shp_custom.site')->validateShortNameUniqueness($short_name);
 }
--- a/web/modules/custom/shepherd/shp_custom/src/Service/Site.php
+++ b/web/modules/custom/shepherd/shp_custom/src/Service/Site.php
@@ -136,7 +136,7 @@ class Site {
    * Load all entities that contain a value in a field.
    *
    * @param string $node_type
-   *   Node type. Defaults to shp_site.
+   *   Node type.
    * @param string $field
    *   Field name.
    * @param string $field_value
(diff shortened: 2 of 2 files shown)


pingers:
Alternatively, I can just create a function in the module to call the orginal service method in a non-static context.

pingers:
After fixing the other issue, happy enough to wrap the service call in a function.

web/modules/custom/shepherd/shp_custom/shp_custom.module
```diff
@@ -299,3 +299,21 @@ function shp_custom_entity_operation_alter(array &$operations, EntityInterface $
...
+ * Ensure that the short_name generated is unique.
+ *
+ * @param string $short_name
+ *   Generated short name.
+ *
+ * @return string
+ *   If not unique adds a number to end of string, otherwise valid.
+ */
+function shp_custom_generate_unique_short_name($short_name) {
+  $results = \Drupal::service('shp_custom.site')->loadEntitiesByFieldValue('shp_site', 'field_shp_short_name', $short_name);
```
fenetikm: This method `loadEntitiesByFieldValue` feels like a more generic piece of functionality that could perhaps just go in a `shp_custom.entity` service.

web/modules/custom/shepherd/shp_custom/shp_custom.module
```diff
@@ -299,3 +299,21 @@ function shp_custom_entity_operation_alter(array &$operations, EntityInterface $
...
+/**
+ * Ensure that the short_name generated is unique.
+ *
+ * @param string $short_name
+ *   Generated short name.
+ *
+ * @return string
+ *   If not unique adds a number to end of string, otherwise valid.
+ */
+function shp_custom_generate_unique_short_name($short_name) {
```
fenetikm: Can't this instead just be a static method on a service / class?

web/modules/custom/shepherd/shp_custom/src/Service/Site.php
```diff
@@ -176,32 +176,14 @@ public function applyJavascriptTitleField(array &$form) {
     $form['title']['widget'][0]['value']['#id'] = 'edit-title';
     $form['field_shp_short_name']['widget'][0]['value']['#type'] = 'machine_name';
     $form['field_shp_short_name']['widget'][0]['value']['#machine_name'] = [
-      'exists' => ['Drupal\shp_custom\Service\Site', 'validateShortNameUniqueness'],
+      'exists' => ['shp_custom_generate_unique_short_name'],
```
fenetikm: Does something like `Drupal\shp_custom\Service\EntityService::validateShortNameUniqueness` work ?

web/modules/custom/shepherd/shp_custom/src/Service/Site.php
```diff
@@ -176,32 +176,14 @@ public function applyJavascriptTitleField(array &$form) {
...
   /**
-   * Ensure that the short_name generated is unique.
-   *
-   * @param string $short_name
-   *   Generated short name.
-   *
-   * @return string
-   *   If not unique adds a number to end of string, otherwise valid.
-   */
-  public function validateShortNameUniqueness($short_name) {
```
fenetikm: if this was static (and probably doesn't belong in this service but whatevs) it would be all good yes?

web/modules/custom/shepherd/shp_custom/shp_custom.module
pingers: Tried that... but the loadEntitiesByFieldValue() would have to be static too...
and that relies on EQ service... so it's not easy.

web/modules/custom/shepherd/shp_custom/src/Service/Site.php
pingers: Nope.. because of the afore mentioned calling non-static methods from a static context.

web/modules/custom/shepherd/shp_custom/src/Service/Site.php
pingers: We could leave it in here and call it via a function wrapper (to fetch the service rather than call the method statically from the callback)

web/modules/custom/shepherd/shp_custom/src/Service/Site.php
```diff
@@ -145,10 +145,10 @@ public function loadRelatedEntitiesByField(Node $node, $reference_field, $node_t
    * @return mixed
    *   Query results.
    */
-  public function loadEntitiesByFieldValue($node_type = 'shp_site', $field, $field_value) {
+  public function loadEntitiesByFieldValue($node_type, $field, $field_value) {
     $results = $this->node->getQuery()
       ->condition('type', $node_type)
-      ->condition($field, $field_value, 'CONTAINS')
+      ->condition($field, $field_value)
```
pingers: Leaving this as it was would cause false positives. Load by exact value.

web/modules/custom/shepherd/shp_custom/src/Service/Site.php
```diff
@@ -189,16 +189,12 @@ public function applyJavascriptTitleField(array &$form) {
    * @param string $short_name
    *   Generated short name.
    *
-   * @return string
+   * @return bool
```
pingers: Must be bool for a uniqueness check.

web/modules/custom/shepherd/shp_custom/src/Service/Site.php
bart88: 👍

bart88:
Happy with this to 🚢

PR #65, opened 2017-11-03
Check that the build has any phase info before making assumptions.

pingers:
On submitting the create environment form, this was causing a php notice.


web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
```diff
@@ -239,7 +239,7 @@ public function createdEnvironment(
     if (!$update_on_image_change) {
       // We need to check if the image is already 'built', or we get an error.
       $build_status = $this->client->getBuilds('', 'buildconfig=' . $build_config_name);
-      if ($build_status['items'][0]['status']['phase'] === 'Complete') {
+      if (count($build_status['items']) && $build_status['items'][0]['status']['phase'] === 'Complete') {
```
bart88: 💯  good catch.

PR #66, opened 2017-11-03
Fix the callback to not supply an object as the callback class and av…

pingers:
…oid calling non-static methods from static context.

This MR should supercede https://github.com/universityofadelaide/shepherd/pull/63. It addresses the actual problem with the callback definition being supplied an object, rather than a class name... which caused more problems because of the static context in which it was called.


bart88:
:shipit:

PR #67, opened 2017-11-06
Feature/webops 258 new entities redirect to view

bart88:
Use the WithDestination class on the action links to add a return destination to the node add forms. 

This improves the usability of Shepherd, creating entities returns to the views from where the user clicked the create/add new buttons.


web/modules/custom/shepherd/shp_custom/src/Plugin/Menu/LocalAction/WithSiteId.php
```diff
@@ -2,15 +2,13 @@
 
 namespace Drupal\shp_custom\Plugin\Menu\LocalAction;
 
-use Drupal\Core\Menu\LocalActionDefault;
 use Drupal\Core\Routing\RouteMatchInterface;
-use Drupal\Core\Routing\RouteProviderInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\group\Plugin\Menu\LocalAction\WithDestination;
```
pingers: Should declare this dependency in the .info.yml.

PR #71, opened 2017-11-27
Provide validation on OpenShift config form.

bart88:
Connect to OpenShift via the client with the user provided values to verify valid configuration. Currently we assume that all config provided to this form is correct and provide no feedback. A user who enters invalid config will not know until they are creating projects or sites and often it fails silently.

### DIFF
--- a/web/modules/custom/shepherd/shp_orchestration/src/Form/OpenShiftConfigEntityForm.php
+++ b/web/modules/custom/shepherd/shp_orchestration/src/Form/OpenShiftConfigEntityForm.php
@@ -152,7 +152,8 @@ class OpenShiftConfigEntityForm extends EntityForm {
   protected function getErrorFieldName(array $client_response): string {
     $field_name = '';
     switch ($client_response['code']) {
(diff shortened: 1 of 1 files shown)


web/modules/custom/shepherd/shp_orchestration/src/Form/OpenShiftConfigEntityForm.php
```diff
@@ -65,6 +67,26 @@ public function form(array $form, FormStateInterface $form_state) {
...
+      $form['namespace']['#value'],
+      $form['verify_tls']['#checked']
+    );
+
+    if (isset($client_response['error']) && $client_response['error']) {
+      $field_name = $this->getErrorFieldName($client_response);
+      $form_state->setErrorByName($field_name, $client_response['code'] . ':' . $client_response['message']);
+    }
+
+    parent::validateForm($form, $form_state);
```
pingers: Usually best practice to do this first.

pingers:
Made some slight changes - but great MR.

PR #76, opened 2017-12-17
Changes to the sites view - more filters

singularo:
Changes	to the sites view - more filters
* Added	domain field and filter
* Changed the date field to be short instead of medium
* Disabled caching due to weirdness with the tags filter field not being updated, and actually that didn't fix it :-(
How do we feel about making that tags filter field an autocomplete for a quick fix?


pingers:
Looking into the tags thing... not working properly for me :confounded:

pingers:
I've completely removed the exposed filter block... not sure why we were doing that. It's not needed and it was the source of the problems... I think because the block filters were interacting with the page filters. *shrugs* Moving on :smile:

PR #80, opened 2018-01-24
Removes the setup flag from the setupVolumes method. Just call genera…

pingers:
…teVolumeData.

Just some clean up and adding todo's.


web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
```diff
@@ -364,6 +364,7 @@ public function promotedEnvironment(
 
     $result = $this->client->updateService($site_deployment_name, $environment_deployment_name);
     if ($result && $clear_cache) {
+      // @todo - Remove drush call, it relates to a project type rather than all projects.
```
bart88: Yes please .. lets remove this in another PR.

bart88:
Thanks for that clarification, happy with this 👍  Lets 🚢 🇮🇹

PR #82, opened 2018-01-26
Add cleanup for created jobs to the delete.

singularo:
Add job delete after cron delete


bart88:
:shipit:

pingers:
If we refactor deleteJobByLabel(), need to update here too.

PR #84, opened 2018-01-30
Feature/webops 288 storage class

pingers:
Adds a storage class name field to projects and passes it through to PVC creation (triggered by environment creation).

To use, just need to add storage class terms to the vocabulary. E.g. bronze, silver etc.


bart88:
Looks good to me 👍   :shipit:

PR #87, opened 2018-02-08
Feature/cron key overlap

pingers:
Fixes duplicate keys causing environments with multiple cron jobs with the same schedules to be overwritten.


davemikolaj:
:ship: :it:

Good example of where tests are required.

PR #90, opened 2018-02-13
RandomStringGenerator tests (and fix)

davemikolaj:
Wrote tests for the class, found some off by one errors and in the unlikely event someone wanted to generate  a 0 character string, they got 2 characters, which annoyed the OCD part of me.

### DIFF
--- a/web/modules/custom/shepherd/shp_custom/src/Service/RandomString.php
+++ b/web/modules/custom/shepherd/shp_custom/src/Service/RandomString.php
@@ -16,11 +16,11 @@ class RandomString {
    *   The length of the string to return.
    *
    * @return string|null
-   *   The generated string. If length supplied is 0, will return null.
+   *   The generated string. If length supplied is 0, will return empty string.
    */
-  public function generate($length = 20) {
+  public function generate(int $length = 20) {
     if ($length === 0) {
-      return NULL;
+      return '';
     }
 
     $count = range(1, $length);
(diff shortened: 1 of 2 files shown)


web/modules/custom/shepherd/shp_custom/src/Service/RandomString.php
```diff
@@ -15,15 +15,20 @@ class RandomString {
...
    *
-   * @return string
-   *   The generated string.
+   * @return string|null
+   *   The generated string. If length supplied is 0, will return null.
    */
   public function generate($length = 20) {
-    $count = range(0, $length);
+    if ($length === 0) {
+      return NULL;
```
pingers: should return empty string? avoid returning different types.

web/modules/custom/shepherd/shp_custom/src/Service/RandomString.php
```diff
@@ -15,15 +15,20 @@ class RandomString {
    * @param int $length
    *   The length of the string to return.
    *
-   * @return string
-   *   The generated string.
+   * @return string|null
+   *   The generated string. If length supplied is 0, will return null.
    */
   public function generate($length = 20) {
```
pingers: missing type hint

davemikolaj: Bah! All you PHP7'ers.

PR #91, opened 2018-02-13
Feature/orchestration exception

pingers:
Incoming...

The orchestration provider not configured exception is evil. It's time to go and replace it with using the dummy provider as the default. We will warn users on the status page to configure and orchestration provider.

Reason(s) for doing this:
* Brittle.
* Causes exceptions at weird times - when there's dependencies that rely on being able to create the orchestration provider during install for example.
* Breaks dependency injection in some cases... so we took shortcuts to work around it using Drupal::service() directly (rather than DI).

### DIFF
--- a/web/modules/custom/shepherd/shp_custom/shp_custom.module
+++ b/web/modules/custom/shepherd/shp_custom/shp_custom.module
@@ -87,30 +87,6 @@ function shp_custom_toolbar_alter(&$items) {
 }
 
 /**
- * Generate a list of projects for a select list.
- *
- * @todo No longer used, delete this.
- *
- * @return array
- *   The choices formatted as id => label.
- */
-function shp_custom_projects() {
-  $ids = \Drupal::entityQuery('node')
-    ->condition('type', 'shp_project')
-    ->sort('title', 'ASC')
-    ->execute();
-
-  $entities = Node::loadMultiple($ids);
-
-  $choices = [];
-  foreach ($entities as $entity) {
-    $choices[$entity->id()] = $entity->getTitle();
-  }
-
-  return $choices;
-}
-
-/**
  * Invalidate site entity caches because environments for sites have changed.
  *
  * @param Drupal\node\NodeInterface $environment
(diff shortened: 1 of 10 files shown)


pingers:
@davemikolaj Do we need that todo for now?

web/modules/custom/shepherd/shp_custom/shp_custom.module
```diff
@@ -132,15 +131,9 @@ function shp_custom_invalidate_site_cache(NodeInterface $environment) {
...
-      /** @var Drupal\shp_orchestration\OrchestrationProviderInterface $orchestration_provider_plugin */
-      $orchestration_provider_plugin = \Drupal::service('plugin.manager.orchestration_provider')
-        ->getProviderInstance();
-    }
-    catch (OrchestrationProviderNotConfiguredException $e) {
-      drupal_set_message($e->getMessage(), 'warning');
-      return FALSE;
-    }
+    /** @var Drupal\shp_orchestration\OrchestrationProviderInterface $orchestration_provider_plugin */
+    $orchestration_provider_plugin = \Drupal::service('plugin.manager.orchestration_provider')
```
fenetikm: all this belongs in an entity service so you can inject it.

fenetikm: or in your case probably an environment / node service I am guessing

davemikolaj: Working on all this bit by bit.. This MR came from me refactoring other things that raised this issue...

pingers: I was thinking about this when I was refactoring. I'm pretty sure the only reason it exists here was to exit early from the hook if there's no orchestration provider configured. So you're probably right about it going away.

web/modules/custom/shepherd/shp_orchestration/shp_orchestration.install
```diff
@@ -0,0 +1,34 @@
...
+
+use Drupal\Core\Url;
+
+/**
+ * Implements hook_requirements().
+ */
+function shp_orchestration_requirements($phase) {
+  $requirements = [];
+
+  if ($phase == 'runtime') {
```
davemikolaj: === ?

PR #93, opened 2018-02-15
Orchestration Provider tests

davemikolaj:
Adds in tests for orchestration provider events and environment events classes.

Also refactor another test I came across to ensure we don't use deprecated functions and add correct doc blocks.


web/modules/custom/shepherd/shp_orchestration/tests/src/Unit/OrchestrationEnvironmentEventTest.php
```diff
@@ -0,0 +1,176 @@
...
+    $this->assertEquals($this->orchestrationProviderMock, $this->orchestrationEvents->getOrchestrationProvider());
+  }
+
+  /**
+   * Test getDeploymentName function.
+   *
+   * @covers ::getDeploymentName
+   */
+  public function testGetDeploymentName() {
+    $this->assertEquals('MyTestDeployment', $this->orchestrationEvents->getDeploymentName());
```
davemikolaj: I'm aware I could have compared the actual mock instead of typing the string in - but it makes it a bit more strict in case both the mock and the return are both null. Same goes with all the other assertions.

web/modules/custom/shepherd/shp_orchestration/tests/src/Unit/OrchestrationEventsTest.php
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * Test the constants.
+   */
+  public function testConstants() {
+    $setup_environment   = 'shp_orchestration.setup_environment';
+    $created_environment = 'shp_orchestration.created_environment';
+    $updated_environment = 'shp_orchestration.updated_environment';
+    $deleted_environment = 'shp_orchestration.deleted_environment';
+
+    $this->assertEquals($setup_environment, $this->orchestrationEvents::SETUP_ENVIRONMENT);
```
pingers: Do we really need to test that constants don't change? I guess that could be a thing.

web/modules/custom/shepherd/shp_orchestration/src/Event/OrchestrationEnvironmentEvent.php
```diff
@@ -26,10 +26,24 @@ class OrchestrationEnvironmentEvent extends Event {
   protected $deploymentName;
 
   /**
-   * Objects related to an environment.
+   * The site this environment is for.
+   *
+   * @var \Drupal\node\Entity\Node
```
fenetikm: should probably use NodeInterface

web/modules/custom/shepherd/shp_orchestration/tests/src/Unit/OrchestrationEnvironmentEventTest.php
```diff
@@ -0,0 +1,176 @@
...
+  /**
+   * Test setEnvironmentVariables and getEnvironmentVariables functions.
+   *
+   * @covers ::setEnvironmentVariables
+   * @covers ::getEnvironmentVariables
+   */
+  public function testEnvironmentVariables() {
+    $this->orchestrationEvents->setEnvironmentVariables($this->testEnvironmentVariables);
+    $this->assertArrayEquals($this->testEnvironmentVariables, $this->orchestrationEvents->getEnvironmentVariables());
+  }
```
fenetikm: missing new line after this.

web/modules/custom/shepherd/shp_orchestration/tests/src/Unit/OrchestrationEventsTest.php
davemikolaj: That's the whole point of tests...

web/modules/custom/shepherd/shp_orchestration/tests/src/Unit/OrchestrationEnvironmentEventTest.php
pingers: Uses more memory, which is more prone to cosmic radiation random fails.

pingers: :sun_with_face: :trollface:

web/modules/custom/shepherd/shp_orchestration/tests/src/Unit/OrchestrationEventsTest.php
```diff
@@ -0,0 +1,49 @@
...
+    $this->assertEquals($created_environment, $this->orchestrationEvents::CREATED_ENVIRONMENT);
+    $this->assertEquals($updated_environment, $this->orchestrationEvents::UPDATED_ENVIRONMENT);
+    $this->assertEquals($deleted_environment, $this->orchestrationEvents::DELETED_ENVIRONMENT);
+  }
+
+}
```
pingers: new line.

PR #94, opened 2018-02-20
Feature/webops 130 user groups

pingers:
Adds Projects, Groups and Sites lists to user profile pages.


davemikolaj:
I'm not a huge fan of this displaying on all /user/* pages - including /edit etc.

Can we look at verifying via code and building only on /user/[id]

davemikolaj:
Also - restrict to Administrators/Online Consultants only? Don't need standard users seeing it.

PR #95, opened 2018-02-23
More dependency injection and Removal of drush site aliases

davemikolaj:
Doing some more quick wins for dependency injection, but it's highlighted a higher level issue of how this project has been structured - we'll all need to have a bit of a workshop session to determine the value of changing a few things.

There's a couple of non-injected services still around, part of the above discussion.

Also removed drush site aliases module - no longer used.


pingers:
This all looks sane. I'll test it tomorrow.

pingers:
All the reviewers! Take a look... but I'm happy with where this got to and everything appears to work in manual testing:
* Backup
* Restore
* Promote to prod
* Links to logs and terminal
* Cron is running

PR #97, opened 2018-03-09
Fixes redis spewing assert() deprecation notices, updates composer de…

pingers:
…ps from scaffold.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,9 +11,11 @@
         "composer/installers":                           "^1.2",
         "cweagans/composer-patches":                     "~1.0",
         "drupal-composer/drupal-scaffold":               "^2.2",
+        "predis/predis":                                 "^1.1.1",
         "universityofadelaide/shepherd-drupal-scaffold": "~1.0",
+        "wikimedia/composer-merge-plugin":               "^1.4",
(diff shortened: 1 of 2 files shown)


davemikolaj:
Seems like updating phpunit has caused drush tests to freak out. Might need to investigate this in another PR.

pingers:
CI fails are due to PHPUnit making a change to the signature of the TestCase run() method as part of the PHPUnit 6 release - expecting a PHPUnit\Framework\TestResult instead of \PHPUnit_Framework_TestResult.

Not sure how it will be fixed in the interim. 
I'm not clear why we need to specify `-c web/core web/module` as part of the CI script.

pingers:
Looks like devel module is most likely to blame and we'll need to lock down phpunit version to match the drush require-dev dependency to solve it in the interim.

davemikolaj:
-c web/core for config
web/modules to specify the directory to be tested.

Surely that module has an issue logged by now, could write a patch for it and all good after that.

pingers:
Won't be a fix until Drush switches over to PHPUnit 6. It's a breaking change - not likely to happen overnight.

PR #101, opened 2018-06-26
Reorder the quickstart commands

bart88:
Mariadb application must be running before you run the `robo content:generate` command. There was  a paragraph below that talked about using the `minishift console` command to access the web ui, lets put that above so it flows better.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -100,11 +100,12 @@ composer install
 robo build
 ```
 
-Login to the OpenShift web ui using the `developer:developer` credentials. If you are using `minishift` you easily access the OpenShift web interface by running: 
+Login to the OpenShift web ui using the `developer:developer` credentials. If you are using `minishift` you can easily access the OpenShift web interface by running: 
+
 ```bash
 minishift console
 ```
-Note : you will recieve an insecure warning in your browser, don't panic. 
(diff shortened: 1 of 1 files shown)


DEVELOPERS.md
```diff
@@ -100,22 +100,24 @@ composer install
...
-robo dev:content-generate
+minishift console
 ```
+*Note: you will receive an insecure warning in your browser, don't panic.* 
 
-Thats it; visit the OpenShift web interface to see a build running and a
-deployment ready to occur when the build finishes. The URL to the web interface
-can be found in the terminal log when Minishift starts. Alternatively, execute
-the following command to open the Minishift console in a browser.
+The `robo build` command will automagically create a project called `myproject` and a deployment of `mariadb` for you.
```
pingers: `deploy.sh` creates the mariadb container.
`robo build` builds and installs the shepherd Drupal distro.
`myproject` is created by OpenShift as a default project when you create a user. We just assume that is the project in things like ShepherdContentGenerate.php to simplify dev. Pretty sure anyway :smile:

bart88: 👍 oh thats right about the myproject. Will update this part of the doc.

singularo:
I think we can probably can this and do a more thorough re-write of the setup.
@pingers

PR #105, opened 2018-08-15
Feature/remove unused vue

pingers:
We'll bring it back if we need it.


bart88:
Agreed lets get rid of it. Progressive decoupling in this instance is not going to work. A decoupled front end should be a future goal.

PR #112, opened 2018-10-15
Feature/webops 339 random cron

pingers:
Generates a random minutes default value to stagger environment cron jobs.
Eg. [shepherd:cron:15] generates a list like "2,17,32,47"


web/modules/custom/shepherd/shp_custom/src/Routing/RouteSubscriber.php
```diff
@@ -21,6 +21,10 @@ protected function alterRoutes(RouteCollection $collection) {
     if ($route = $collection->get('view.shp_site_users.page_1')) {
       $route->setRequirement('_custom_access', '\Drupal\shp_custom\Controller\SiteLocalTaskController::checkAccess');
     }
+
+    if ($route = $collection->get('')) {
+      $route->setDefault('_form', '\Drupal\form_overwrite\Form\NewUserLoginForm');
+    }
```
acbramley: What does this do?

pingers: Good point - wasn't meant to be committed.

singularo:
Shouldn't this be merging to develop @pingers ?

PR #114, opened 2018-10-25
Fixes promote environment production url missing path bug.

pingers:


singularo:
Whoops, easy fix, but.

PR #116, opened 2018-11-27
Archived state should be a default revision

acbramley:
Without this, https://github.com/universityofadelaide/shepherd/blob/master/web/modules/custom/shepherd/shp_custom/shp_custom.module#L123 will always back the environment up since the revision loaded will be the published version.


pingers:
:ship:

PR #118, opened 2019-01-09
Changes so that we're using the correct version of Robo and other utils for building

singularo:
shepherd-drupal-scaffold ~1.0 didn't have drush cimy, so the build wouldn't work.
Add in dsh.project and changes to suit https://github.com/universityofadelaide/shepherd-drupal-scaffold/pull/51
Add in openshift startup/shutdown util script (seperate to dsh as that is more generic).


docker-compose.linux.yml
```diff
@@ -0,0 +1,47 @@
+version: '3'
+services:
+  web:
+    image: singularo/apache2-php7-dev:shepherd
```
singularo: This should be the university one.

openshift
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+# `set +e` is used to continue on errors throughout this script.
+set -euo pipefail
+IFS=$'\n\t'
+
+BASE_DIR="/openshift/uni-shepherd"
```
singularo: Hmm this might not be a good idea, maybe /var/ or /opt/ would be nicer.

PR #121, opened 2019-01-31
Feature/detach scaffold

singularo:
Detach scaffold
Support linux properly again
Support new redis config


composer.json
```diff
@@ -44,15 +44,22 @@
             "php": "7.2.0"
         }
     },
+    "autoload": {
+        "classmap": [
+            "scripts/composer/ScriptHandler.php",
+            "scripts/composer/Plugin.php"
```
pingers: This is about how we override the scaffold... to have a shepherd specific dsh.
composer update is still overwriting dsh. I think I can refactor the scaffold to fix by having a method to provide the list of files which are copy-if-doesnt-exist and copy-always.

singularo: Do we want to worry about that with this PR?

pingers: Nope... I think it's fine... remembered that it's too hard for now because there's 2 subscribers to that event... which makes it tricky to override.

PR #123, opened 2019-02-04
Confirm logged in as developer or die, nicer notices, nicer if's.

singularo:


dsh
```diff
@@ -74,13 +74,13 @@ dsh_openshift_init() {
...
+    notice "Using OpenShift in Docker."
   fi
 
-  # Are we logged in?
-  if [ ! $(oc whoami) ]; then
-    # Prompt for login.
-    oc login
+  # Are we logged in as local developer, if not, bail!
+  if [[ $(oc whoami) != "developer" ]]; then
+    error "You are not logged in as developer, exiting for safety reasons."
```
pingers: I'd add "Please login to your local development OpenShift."

singularo: :+1: Done

dsh
```diff
@@ -92,24 +92,26 @@ dsh_openshift_init() {
       oc annotate svc mysql-external "service.alpha.openshift.io/dependencies=[{\"name\": \"mariadb\", \"kind\": \"Service\"}]"
     fi
   else
-    echo "Not using Minishift, assuming mysql is available on port 3306."
```
pingers: Above this is a service check for mysql - could just `oc get svc mysql`?

singularo: Probably, don't have minishift to test, but.

pingers:
Should probably fix the dsh_shell docker ps check... it changed upstream.
`if docker-compose -f ${DOCKER_COMPOSE_FILE} ps --filter "status=stopped" --services | grep web > /dev/null; then`

pingers:
dsh status changed too...
https://github.com/universityofadelaide/shepherd-drupal-scaffold/blob/master/dsh#L83

pingers:
:ship:

PR #128, opened 2019-02-18
Feature/webops 302

singularo:
Some doc updates
Incorporates dont crash on invalid backup string fix.
Incorporates dont mount backup volume for web containers


singularo:
Removed the password stuff, that can be done separately.

singularo:
Hmm good question. Manuel I think.

pingers:
manuel

PR #129, opened 2019-02-19
Remove online consultant role from Shepherd distro.

pingers:
This in a university specific role that can be preserved in the ua_shepherd repo (by merging and reverting this commit in that repo).


singularo:
Seems legit. Built ok created env etc.
:ship: it out..

PR #130, opened 2019-02-28
Feature/config override secrets

pingers:
Adds project defaults and environment specific configuration overrides via secrets.

Secrets get mounted in /etc/secret-environment and /etc/secret-project - it's then up to the applications to read and respect precedence of those secrets.

### DIFF
--- a/web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
+++ b/web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
@@ -229,7 +229,7 @@ class OpenShiftOrchestrationProvider extends OrchestrationProviderBase {
     // Tell, don't ask (to create a build config).
     $this->createBuildConfig($build_config_name, $source_ref, $source_repo, $builder_image, $source_secret, $image_stream_tag, $formatted_env_vars);
 
-    if (!$volumes = $this->setupVolumes($project_name, $deployment_name, $storage_class)) {
+    if (!$volumes = $this->setupVolumes($project_name, $deployment_name, $storage_class, $secrets)) {
       return FALSE;
     }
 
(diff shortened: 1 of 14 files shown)


web/modules/custom/shepherd/shp_orchestration/src/Service/Configuration.php
```diff
@@ -128,7 +128,21 @@ public function extractEnvironmentVariables(NodeInterface $node) {
    *   An array of secrets.
    */
   public function getSecrets(NodeInterface $node) {
-    return $this->moduleHandler->invokeAll('shp_secrets', [$node]);
```
pingers: Removed this as there's no implementations and doesn't make a lot of sense now that we can just attach secrets to deployments.

singularo: :+1:

web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
```diff
@@ -324,7 +325,7 @@ public function updatedEnvironment(
     $deployment_name = self::generateDeploymentName($environment_id);
     $formatted_env_vars = $this->formatEnvVars($environment_variables, $deployment_name);
 
-    if (!$volumes = $this->setupVolumes($project_name, $deployment_name)) {
+    if (!$volumes = $this->setupVolumes($project_name, $deployment_name, $storage_class, $secrets)) {
```
singularo: That was missing? weird.

singularo: storage class, I mean

pingers: Ya, go figure.

PR #131, opened 2019-04-03
Delay creation of production route until a site is promoted.

singularo:
- Add in extra required parameters to the Interface, Dummy and Openshift providers
- Change promotion to:
  - Check if the service/route already exists
  - If not, create both
- Leave site creation still doing the call to the provider
- Remove creation of service/route when site is created from the provider level
- Move the annotations adjustments and domain/path to the promoted call

### DIFF
--- a/web/modules/custom/shepherd/shp_orchestration/src/OrchestrationProviderInterface.php
+++ b/web/modules/custom/shepherd/shp_orchestration/src/OrchestrationProviderInterface.php
@@ -296,13 +296,11 @@ interface OrchestrationProviderInterface extends PluginInspectionInterface {
    *   The domain name of the site.
    * @param string $path
    *   The path of the site.
-   * @param array $annotations
-   *   An array of route annotations.
    *
    * @return bool
    *   Returns true if succeeded.
    */
-  public function createdSite(string $project_name, string $short_name, int $site_id, string $domain_name, string $path, array $annotations = []);
+  public function createdSite(string $project_name, string $short_name, int $site_id, string $domain_name, string $path);
 
   /**
    * Handles a site being updated.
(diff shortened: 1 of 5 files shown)


web/modules/custom/shepherd/shp_orchestration/src/Service/Environment.php
```diff
@@ -306,11 +306,24 @@ public function promoted(NodeInterface $site, NodeInterface $environment, bool $
       return FALSE;
     }
 
+    // Load the taxonomy term that has protect enabled.
+    $promoted_term = $this->environmentType->getPromotedTerm();
+
+    // Extract and transform the annotations from the environment type.
+    $annotations = $promoted_term ? $promoted_term->field_shp_annotations->getValue() : [];
```
pingers: I think we should just remove the stale Site::created() equivalent code for annotations. Custom orchestration providers can figure out annotations themselves as needed.

web/modules/custom/shepherd/shp_orchestration/src/OrchestrationProviderInterface.php
```diff
@@ -296,13 +296,11 @@ public function promotedEnvironment(
...
    * @param string $path
    *   The path of the site.
-   * @param array $annotations
-   *   An array of route annotations.
    *
    * @return bool
    *   Returns true if succeeded.
    */
-  public function createdSite(string $project_name, string $short_name, int $site_id, string $domain_name, string $path, array $annotations = []);
+  public function createdSite(string $project_name, string $short_name, int $site_id, string $domain_name, string $path);
```
pingers: I think this got missed in the dummy provider?

pingers:
:shipit:

PR #132, opened 2019-04-05
Queue improvements

singularo:
- If a job can't be found in openshift, assume its done and carry on.
- Add some warnings when processing jobs to make diagnosis easier.

### DIFF
--- a/web/modules/custom/shepherd/shp_backup/src/Plugin/QueueWorker/BackupQueueWorker.php
+++ b/web/modules/custom/shepherd/shp_backup/src/Plugin/QueueWorker/BackupQueueWorker.php
@@ -17,13 +17,18 @@ class BackupQueueWorker extends BackupQueueWorkerBase {
    */
   public function processItem($job) {
     // Load the node for the job.
+    /** @var $backup \Drupal\node\Entity\Node */
     if ($backup = $this->nodeStorage->load($job->backupId)) {
       // Perform the job.
       if ($responseBody = $this->backup->create($backup)) {
         // Update the job name for isComplete() check.
         $this->setJobName($job, $responseBody);
+        return;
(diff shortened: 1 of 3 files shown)


web/modules/custom/shepherd/shp_backup/src/Plugin/QueueWorker/BackupQueueWorker.php
```diff
@@ -17,13 +17,18 @@ class BackupQueueWorker extends BackupQueueWorkerBase {
    */
   public function processItem($job) {
     // Load the node for the job.
+    /** @var $backup \Drupal\node\Entity\Node */
```
pingers: Interface

web/modules/custom/shepherd/shp_backup/src/Plugin/QueueWorker/BackupQueueWorker.php
```diff
@@ -17,13 +17,18 @@ class BackupQueueWorker extends BackupQueueWorkerBase {
    */
   public function processItem($job) {
     // Load the node for the job.
+    /** @var $backup \Drupal\node\NodeInterface */
```
pingers: Single asterisk... double asterisk is for multiline comments.

pingers:
Very minor fix, then :ship:

web/modules/custom/shepherd/shp_backup/src/Plugin/QueueWorker/BackupQueueWorker.php
singularo: No. no @pingers phpstorm does @var etc automatically after the space after /** but not after /* and everywhere else is /** @var etc..

PR #135, opened 2019-04-25
Change backup PVC logic.

singularo:
The recent change that I did to only create the backup PVC for jobs
has a problem. The executeJob function doesn't have the info to
create the PVC, so it can never be created.

The PVC needs to be created, but only mounted as a volume as required.

This is changed with this commit.


pingers:
Err :hankey:

Trying to understand this bug has just uncovered another bug with the project and environment secrets - they don't get attached to jobs. I.e. the prj and env secrets should be declared in generateVolumeData().

singularo:
Moved the secrets and adjusted a bit. We can probably merge this, then look at fixing passing in the array of secrets in the executeJob etc as a separate thing as we clearly haven't needed that yet ;-/

PR #136, opened 2019-05-01
Don't generate operations links when entities have been no longer exist.

pingers:


singularo:
Looks good to me, stops the issue from occurring.

PR #137, opened 2019-05-08
More changes in the backup logic.

singularo:
Previous update didn't _quite_ fix the problem.
- The PVC _needs_ to always be created.
- The backup name creation needs to be outside of generateVolumes() so that the pvc can be created.
- The generateVolumes() function is changed slightly to take the constructed names, rather than building inside the function.


singularo:
No. rethink whole thing.

PR #138, opened 2019-05-09
Feature/backup pvc issue

singularo:
- Move the volume name generation to its own function.
- Call the generator to get the volume names.
- Keep mounting logic separate.
- Fix multiple cron's not working issue.


web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
```diff
@@ -259,6 +261,8 @@ public function createdEnvironment(
       return FALSE;
     }
 
+    // Get the volumes and include the backups this time.
+    $volumes = $this->generateVolumeData($project_name, $deployment_name, $secrets, TRUE);
```
pingers: I'd just assign to another var... avoid mutation.

pingers:
One fix, then :ship:

PR #139, opened 2019-05-24
Feature/add memory cpu support

singularo:
Add new fields for cpu & memory request & limits.
Put fields on project and site (if environment specific is required, it can just be done through openshift).
Put new fields in a collapsed group by default.
Load the values and add to the data for deploying.


config-export/core.entity_form_display.node.shp_site.default.yml
```diff
@@ -3,23 +3,62 @@ langcode: en
 status: true
 dependencies:
   config:
+    - field.field.node.shp_site.field_shp_cpu_limit
```
pingers: All the other settings like this are on project & environment... which allows a bit more control over things like dev vs prod. Thoughts?

web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
```diff
@@ -847,14 +855,52 @@ protected function formatEnvVars(array $environment_variables, string $deploymen
...
+
+    $fields = [
+      'field_shp_cpu_request'    => 'cpu_request',
+      'field_shp_cpu_limit'      => 'cpu_limit',
+      'field_shp_memory_request' => 'memory_request',
+      'field_shp_memory_limit'   => 'memory_limit',
+    ];
+
+    foreach ($fields as $field => $client_field) {
+      if (!empty($site->{$field}->value)) {
```
pingers: !$site->{$field}->isEmpty()

web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
```diff
@@ -847,14 +855,52 @@ protected function formatEnvVars(array $environment_variables, string $deploymen
...
   /**
-   * Format an array of deployment data ready to pass to OpenShift.
+   * @param int $site_id
+   *   The ID of the site the environment represents.
    *
-   * @todo - move this into the client?
+   * @return array
+   *   Array with cpu & memory request & limits.
+   */
+  protected function setupRequestLimits(int $site_id) {
```
pingers: s/setupRequestLimits/generateRequestLimits

This doesn't setup anything, which is different to setupVolumes(), it's more like generateDeploymentName(), just generating an API specific array structure.

pingers:
There's a gotcha with requests vs limits, so it might be good to add some help text to the fields. I.e. The limit cannot be less than the request.

web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
singularo: Yeah, I was thinking getRequestLimits, but thought that can wait for review ;-)

config-export/core.entity_form_display.node.shp_site.default.yml
singularo: I dunno. I feel like if an environment starts to give issues, you are going to either increase it at the site level (eg particularly large site), or project (new module has pushed things higher).

Environment is probably more when troubleshooting, and likely to be done directly in OpenShift and then adjusted in one of the higher ups depending on what the reason is.

That was my thought process, but I'm not insistent either way, easy enough to change.

web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
singularo: generate it is.

web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
singularo: Nicer.

config-export/core.entity_form_display.node.shp_site.default.yml
singularo: Only this one left to decide. I think my logic is sound, change my mind ;-)

config-export/shp_orchestration.openshift.openshift.yml
```diff
@@ -0,0 +1,11 @@
+uuid: fcd70570-09ac-4a93-8159-61cdd7fc9122
+langcode: en
+status: true
+dependencies: {  }
+id: openshift
+endpoint: 'https://172.17.0.1:8443'
+token: eyJhbGciOiJSUzI1NiIsImtpZCI6IiJ9.eyJpc3MiOiJrdWJlcm5ldGVzL3NlcnZpY2VhY2NvdW50Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9uYW1lc3BhY2UiOiJteXByb2plY3QiLCJrdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3NlY3JldC5uYW1lIjoic2hlcGhlcmQtdG9rZW4tOXg1ZmIiLCJrdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3NlcnZpY2UtYWNjb3VudC5uYW1lIjoic2hlcGhlcmQiLCJrdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3NlcnZpY2UtYWNjb3VudC51aWQiOiJlMWU0Y2MxNi04MDExLTExZTktODEzNy00ODJhZTMyN2Y0MGQiLCJzdWIiOiJzeXN0ZW06c2VydmljZWFjY291bnQ6bXlwcm9qZWN0OnNoZXBoZXJkIn0.CEbNxLYcvf7kVd_zYoQNRUSw-IC5nfkFVUTMMhCxqgd5DZEE_KhYV6Vt235vKSIbt1re2HaVifMryyGYoyZOYqap29vfeS095cD9d5fKAMf2D3YIi7yHvJDBKZhjY7PS62bptONwtkwVYHn0iMJBC3GsSmRv6xUqDyCQAyKgnNKUvm3fq838m1RM6pCH0Yg-SJkKXuN9qjrqbGdmdaniL7sgPY7ZyA2h68OIsvXm_ENv26VmI5DCafzHmfdGbBZAuVbRk_Kuaeh7pYzOBVT2jpCg2Ygb5hKhqmY9vNqcTM_yCjDizEiYs3DfsD9sO0nvI1blLKSn1nMP5c_oxAj45Q
```
pingers: Uh?

singularo: Removed, and added to exclusions.

singularo: Fixed.

config-export/core.entity_form_display.node.shp_site.default.yml
pingers: Well, there's other things like:
* OpenShift DC is where it lives, which is closer to an environment - if we head towards OpenShift as source of truth, smoother transition.
* Site level means cascading updates to environments.

change my mind

singularo: ok, convinced me. changed.

singularo:
Ok. All ready to go now I believe.

pingers:
Just move env request/limit fields into Advanced fieldset and :ship:

PR #140, opened 2019-06-12
Fix minishift conditions and add OSX files to ignore

acbramley:


RoboFile.php
```diff
@@ -16,14 +17,6 @@
  */
 class RoboFile extends RoboFileBase {
 
-  /**
-   * {@inheritdoc}
-   */
-  public function __construct() {
-    parent::__construct();
-    // Put project specific overrides here, below the parent constructor.
-  }
```
pingers: This is just default "docs" from scaffold.

PR #141, opened 2019-07-04
Support single redis host.

singularo:
If we go down the https://www.keydb.dev/ path, this will make things easier.
Just requires manually setting the redis_host in the config and then subsequent deploy's will skip deploying redis containers.
Existing deploys rolled over will still remove the old redis containers
So moving to this will be fairly painless.

But will depend on infinispan etc as to whether needed.


web/modules/custom/shepherd/shp_redis_support/shp_redis_support.module
```diff
@@ -18,8 +18,17 @@ function shp_redis_support_shp_env_vars(NodeInterface $environment) {
   if ($orchestration_provider->getPluginId() === 'openshift_with_redis') {
     $deployment_name = $orchestration_provider::generateDeploymentName($environment->id());
     $env_vars['REDIS_ENABLED'] = '1';
-    $env_vars['REDIS_HOST'] = $deployment_name . '-redis';
     $env_vars['REDIS_PREFIX'] = $deployment_name;
+
+    // Allow for a single large redis host
+    $config_factory = \Drupal::service('config.factory');
+    if ($redis_host = $config_factory->get('shp_orchestration.settings')->get('redis_host')) {
```
pingers: Do we need !empty() check here?

pingers: Also probably can refactor to 2 lines:
```php
$redis_host = $config_factory->get('shp_orchestration.settings')->get('redis_host');
$env_vars['REDIS_HOST'] = $redis_host ?? $deployment_name . '-redis';
```

singularo: Ayup.

PR #143, opened 2019-07-09
Restic backup and restores

acbramley:
**This PR does the following:**
- Rips out all of the Backup logic for creating Backup nodes (still need to figure out what to do with existing Backup nodes and configuration)
- Adds direct communication with Openshift in order to issue Backup and Restore operations, creating the CRs, listing them in custom controllers, etc.
- Adds Backup schedule field to Environment type taxonomy in order to schedule backups for all environments of a type. This means there's 1 config to schedule all production environments to backup nightly (for example). This could be extended to add an override at the environment level etc quite easily.
- Adds labels to all objects that relate to an environment
- Deletes all the queue worker stuff around backups
- A bunch of minor refactoring and tidy ups, protecting against deleting sites etc.

Relies on https://github.com/universityofadelaide/openshift-client/pull/37

**Benefits of this:**
- Drupal no longer holds the state of the backups, meaning we have much more flexibility in exposing information about the backups/restores. E.g on the restore form you can now only selected Backups which have completed successfully.
- We can track the progress of a backup/restore through the UI (the Phase)
- Abstracts logic for backing up/restoring/syncing things into more manageable chunks (no more scary bash scripts in Drupal config!!!)

**This PR doesn't do the following:**
Testing...again with the lack of testing infrastructure set up to allow mocking of clients etc it's very difficult to test this functionality and I simply ran out of time.


composer.json
```diff
@@ -55,6 +55,9 @@
         "post-install-cmd": [
             "@composer drupal:scaffold",
             "UniversityOfAdelaide\\Shepherd\\Plugin::scaffold"
+        ],
+        "post-update-cmd": [
+            "echo \"Checking out dsh that scaffold overrides\"; git checkout dsh"
```
pingers: :it: :ok_hand:

PR #145, opened 2019-07-30
S2i dev image

pingers:
Ran through the full install and dev content generate... seems fine.

Makes it consistent with ua-wcms-foundation s2i docker image.

### DIFF
--- a/docker-compose.linux.yml
+++ b/docker-compose.linux.yml
@@ -45,6 +45,8 @@ services:
 
   db:
     image: mariadb
+    ports:
+      - 3306:3306
     environment:
       MYSQL_DATABASE: drupal
       MYSQL_USER: user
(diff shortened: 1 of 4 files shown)


singularo:
Ahh excellent. It was on my mental 'todo'.

singularo:
You missed the docker/Dockerfile, which i reckon we should have in here, as the ua-wcms-foundation one is in the UA private repo.

singularo:
Oh, and the xdebug.ini for that.

singularo:
Tested via:
cd /tmp
git clone git@github.com:universityofadelaide/shepherd.git
cd shepherd
composer install
./openshift start
./dsh
robo build
robo dev:drupal-content-generate

Seems to work fine, lets get it in and fix anything also after.

PR #146, opened 2019-08-08
Catch the edge case when submitting a new project

singularo:
Errors printed by DrupalContentGenerate

* Change the default env domain name to be as it would be auto generated.
* Add phpcodesniffer auto config module.
* If there is no image stream yet:
  - don't try and get the value of the array
  - do give a warning when in the UI, content generate script is just back to silent now.

### DIFF
--- a/web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
+++ b/web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
@@ -273,23 +273,27 @@ class OpenShiftOrchestrationProvider extends OrchestrationProviderBase {
 
     // Retrieve image stream that will be used for this site. There is only a
     // tiny chance it will be different to the deployment config image.
-    $image_stream = $this->client->getImageStream($sanitised_project_name);
-
-    // Look through the image stream tags to find the one being deployed.
-    foreach ($image_stream['status']['tags'] as $index => $images) {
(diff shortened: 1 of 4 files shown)


web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php
```diff
@@ -273,24 +273,28 @@ public function createdEnvironment(
...
-        if ($image = $images['items'][0]['dockerImageReference'] ?? FALSE) {
-          $this->createCronJobs(
-            $deployment_name,
-            $cron_suspended,
-            $cron_jobs,
-            $image,
-            $cron_volumes,
-            $deploy_data
-          );
+    if (($image_stream = $this->client->getImageStream($sanitised_project_name))
```
pingers: Just do the assign first to clean up the if.

pingers: Then :ship:

singularo: Done

singularo:
Actually, script does output the message.
```
🐳 simon@shepherd:/code$ robo dev:drupal-content-generate
 [Exec] Running bin/drush scr DrupalContentGenerate.php --uri=shepherd.172.17.0.1.xip.io
 [notice] Message: Build not yet complete, cron jobs cannot be created yet.

 [Exec] Done in 3.016s
```

PR #148, opened 2019-08-23
Make the database provisioner commands configurable in the UI.

singularo:
This should really be moved to being a per project style command, but a lot of the code will be basically the same, so.. here tis!


singularo:
Caught up to current develop

PR #149, opened 2019-08-29
Backport backup volume and cron image fixes from develop.

singularo:
Tested with 3.9 and 3.11 and builds etc ok.
Also includes terminal & log fixes.


singularo:
Re-worked this with cherry-pick (the correct way), and re-tested with 3.11, looks like we'll be deploying after 3.9 is gone anyway as thats happening this week.

Did full build from scratch, installed umami (some timeouts during install, normal), performed backup and restore using drush shp-p in the shepherd container, all seemed to work ok.

Removed deployed site, environment deleted as expected.

PR #151, opened 2019-11-25
Feature/sapp 258 max user connections

pingers:
Refactor and add some tests for the database provisioner - along with support for resource limits for provisioned users.

Still not great, but likely better than before.


web/modules/custom/shepherd/shp_database_provisioner/shp_database_provisioner.post_update.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+/**
```
singularo: I didn't have an older database to test with, but looks right.

web/modules/custom/shepherd/shp_database_provisioner/src/Service/Provisioner.php
```diff
@@ -0,0 +1,282 @@
...
+   * @param \mysqli $db
+   *   Privileged database connection.
+   *
+   * @return bool
+   *   True on success otherwise false.
+   */
+  public function createDatabase(string $database, mysqli $db): bool {
+    $query = sprintf('CREATE DATABASE `%s`', $database);
+    $statement = $db->prepare($query);
+    if ($statement === NULL) {
```
singularo: Shouldn't this check for FALSE as well, like added to the createUser.

web/modules/custom/shepherd/shp_database_provisioner/src/Service/Provisioner.php
```diff
@@ -0,0 +1,282 @@
...
+   * @param \mysqli $db
+   *   Privileged database connection.
+   *
+   * @return bool
+   *   True on success otherwise false.
+   */
+  public function dropDatabase(string $database, mysqli $db): bool {
+    $query = sprintf('DROP DATABASE `%s`', $database);
+    $statement = $db->prepare($query);
+    if ($statement === NULL) {
```
singularo: And here.

web/modules/custom/shepherd/shp_database_provisioner/src/Service/Provisioner.php
```diff
@@ -0,0 +1,282 @@
...
+  public function dropUser(
+    string $username,
+    mysqli $db
+  ): bool {
+    $query = sprintf(
+      "DROP USER `%s`@`%%`",
+      $username
+    );
+    $statement = $db->prepare($query);
+    if ($statement === NULL) {
```
singularo: Again

PR #154, opened 2020-01-23
Create, update or delete environment backup schedule according to env…

pingers:
…ironment type.

Using `oc get backupscheduleds` to show the schedule, I've tested the following conditions:
* Single environment promote to production :heavy_check_mark:
* Single environment demote (edit env and set to dev) :heavy_check_mark:
* Multiple environment, promote env2 to prd and auto demote env1 to dev  :heavy_check_mark:

Requires https://github.com/universityofadelaide/openshift-client/pull/40

### DIFF
--- a/web/modules/custom/shepherd/shp_orchestration/src/Service/Environment.php
+++ b/web/modules/custom/shepherd/shp_orchestration/src/Service/Environment.php
@@ -240,6 +240,9 @@ class Environment extends EntityActionBase {
       $storage_class = Term::load($project->field_shp_storage_class->target_id)->label();
     }
 
+    $environment_type = Term::load($node->field_shp_environment_type->target_id);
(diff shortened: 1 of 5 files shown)


web/modules/custom/shepherd/shp_orchestration/src/Service/Environment.php
```diff
@@ -347,7 +352,6 @@ public function promoted(NodeInterface $site, NodeInterface $environment, bool $
     // @todo everything is exclusive for now, implement non-exclusive?
     // Load a non protected term.
     $demoted_term = $this->environmentType->getDemotedTerm();
-    $promoted_term = $this->environmentType->getPromotedTerm();
```
pingers: This was being set twice for no good reason (see L326).

singularo:
Ship this one too. Ship all the things.

PR #155, opened 2020-03-11
Added MINISHIFT_DRIVER variable to configure vm type.

nicksantamaria:
Fixes an issue with the xhyve driver when running virtualbox 6.x.x

```
Error starting the VM: Error creating the VM. Error with pre-create check: "Virtual Box version 4 or lower will cause a kernel panicif xhyve tries to run. You are running version: 6.1.4r136177\n\n\t Please upgrade to version 5 at https://www.virtualbox.org/wiki/Downloads"
```

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -54,7 +54,7 @@ minishift config set memory 4096
 minishift config set cpus 4
 ```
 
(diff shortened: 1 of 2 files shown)


pingers:
Nice one!

PR #163, opened 2020-11-17
SAPP-1082: Adds local volume mount

acbramley:
Adds an emptyDir mount to `/local` for things like twig caches.

Requires projects to update scaffold to include https://github.com/universityofadelaide/shepherd-drupal-scaffold/pull/88/files (will need a new release for WCMS/Admission/Scholarships).

Can be tested with https://github.com/universityofadelaide/shepherd-example-drupal

Environment/Project config must pass the following environment variable:
```
LOCAL_DIR=/local
```


acbramley:
Failures due to composer 2 issues.

pingers:
Okay, we reaaaally need to fix that composer version.

pingers:
@acbramley At least it's failing on linting now ;)

pingers (approved):
Updates openshift lib and declares the empty mount type - looks good to me :+1:

PR #169, opened 2021-03-29
Sapp 1428 https routes

pingers:
Move to using route object in the client and sets https config on all routes to Allow with edge termination.

Related to https://github.com/universityofadelaide/openshift-client/pull/49


singularo:
Looks good to me!

PR #177, opened 2021-09-02
Set cache backend from field default.

pingers:
Currently, the source environment dictates the cache backend. This isn't ideal, because we're trying to move to a different backend. This PR uses the cache backend field default value to populate the new environment cache backend.


singularo:
Thanks for this, yeah, def want to force the switch.
Looks good :shipit:

PR #182, opened 2022-03-30
Upgrade to Drupal 9

singularo:


singularo:
@pingers You should now be able to:
```
./openshift start
./dsh
robo build
```

So d9 is basically working, more to do on talking to microshift though.

singularo:
Bah.. microshift might not work after all - https://github.com/redhat-et/microshift/issues/633

pingers:
I think the config_ignore issue... and the blocks are resolved now.. need to test.

pingers:
This is probably just about ready to go - the only thing I want to test is:

1. fresh build on develop (D8)
2. drupal content generate
3. switch to this branch and run updates
4. check things over

I'm satisfied that things are otherwise working.

pingers:
I'm not sure about the block config. Blocks are ignored by config management, so they shouldn't get deleted. Not sure why you're having issues with that.

singularo:
I think its because the config_ignore.yml in the shepherd profile is not the same as in the imported config, so first time import, it doesn't ignore them.

docker-compose.osx.yml
```diff
@@ -73,18 +73,7 @@ services:
...
-      APP_PASS: pass
-      MGMT_USER: admin
-      MGMT_PASS: admin
-    entrypoint:
-      - docker-entrypoint.sh
-      - -c
-      - standalone-memcached.xml
-    volumes:
-      - ./docker/standalone-memcached.xml:/opt/jboss/infinispan-server/standalone/configuration/standalone-memcached.xml
+    image: redis:alpine
```
pingers, line 76: memcached:alpine???

PR #183, opened 2022-04-12
Switch from OpenShift 3.x to 4.x support

singularo:
This branch is now ready for review by git cloning, then.
```
./openshift start
robo build
robo DrupalContentGenerate.php
```

Known issues:
* datagrid image doesn't get pulled down.
* somethings not quite right with connecting to mysql for the generated environments.


singularo:
Ahh the database isn't being provisioned atm. looking into that.

singularo:
This should be ready for review/testing @pingers. Actual changes aren't that many, I'll start fixing the tests on the client changes now so that can be merged as well.

pingers:
Can we just leave this branch open for now as the integration branch?
(Close this PR and further 4.x work from this branch)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,88 @@
+#
+# GitHub workflow for Shepherd testing.
```
pingers, line 2: Cool, where did you find this? Or had done something similar already?

DEVELOPERS.md
```diff
@@ -28,10 +27,7 @@ Not similar to a production deployment:
 ## Prerequisites
 
 ### All systems
-* Either [Minishift](https://www.openshift.org/minishift/) virtual machine - which
-  provides an OpenShift cluster for local development or
-* [OpenShift](https://github.com/openshift/origin/blob/release-3.11/docs/cluster_up_down.md)
-  running in docker - rather than in a virtual machine.
+* [Code Ready Containers](https://developers.redhat.com/products/openshift-local/overview)
```
pingers: Maybe make a point here about it being renamed to openshift local?

composer.json
```diff
@@ -25,10 +25,11 @@
         "drupal/r4032login": "2.*",
         "drupal/route_condition": "^2.0",
         "drupal/token": "1.*",
+        "drupal/unused_modules": "^1.3",
```
pingers: Shouldn't this only be during dev?

.github/workflows/ci.yml
singularo: Borrowed from elsewhere, adapted for here, using myself elsewhere. Simpler than the multi-container way. Open Source FTW.

DEVELOPERS.md
singularo: Changed the name to the new one, we'll just ignore that it used to be called something else.

composer.json
singularo: Yer. moved.

PR #184, opened 2022-05-18
WIP: sapp 2505 - multiple service accounts

singularo:
With OpenShift 4.x, we will need to use multiple service accounts, each with its own quota.
This MR adds:
A config entity for the service account and the associated crud things.
A service to retrieve a random account from the enabled accounts.


pingers:
Probably want this config in config_ignore

web/modules/custom/shepherd/shp_orchestration/src/OpenShiftClientFactory.php
```diff
@@ -28,9 +28,18 @@ public function __construct(ConfigFactoryInterface $configFactory) {
    */
   public function getClient() {
```
singularo, line 29: This is a bit problematic as its using a factory which doesn't take parameters in Drupal, but for subsequent requests, shepherd needs to store and know which service account/token was used. So while this random token proves it works, shepherd really needs to work that out before creating the client.

So each call that uses the OpenShiftClientFactory will need to know the token to use beforehand and pass it in.
We can make a setToken method here, but it will need to create a new OpenShiftClient, sooo seems a bit strange.

I guess I'm questioning the value of the factory now as there will be differences at call time.

web/modules/custom/shepherd/shp_orchestration/src/OpenShiftClientFactory.php
```diff
@@ -28,9 +28,18 @@ public function __construct(ConfigFactoryInterface $configFactory) {
...
     $settings = $this->configFactory->get('shp_orchestration.settings');
+    try {
+      $serviceAccount = \Drupal::service('shp_service_accounts.random')
```
behzad1q: inject the service

web/modules/custom/shepherd/shp_service_accounts/src/Form/ServiceAccountForm.php
```diff
@@ -0,0 +1,77 @@
...
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function save(array $form, FormStateInterface $form_state) {
+    $result = parent::save($form, $form_state);
+    $message_args = ['%label' => $this->entity->label()];
+    $message = $result == SAVED_NEW
```
behzad1q, line 69: is \SAVED_NEW works too?

web/modules/custom/shepherd/shp_service_accounts/src/Service/ServiceAccountRandomSelection.php
```diff
@@ -0,0 +1,59 @@
...
+   *   The ServiceAccount or NULL
+   */
+  public function getServiceAccount(): ?ServiceAccount {
+    $serviceAccountList = $this->entityTypeManager->getStorage('service_account')->loadByProperties([
+      'status' => TRUE,
+    ]);
+
+    $entries = count($serviceAccountList);
+    if ($entries == 0) {
+      throw new \Exception("No service accounts defined.");
```
behzad1q: use a better Exception, not the general one. MissingRequiredValueException or RequiredDataMissingException. and then define your excpetion class with custom message. here is an example. https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Plugin%21Exception%21MissingValueContextException.php/class/MissingValueContextException/8.8.x

DrupalContentGenerate.php
```diff
@@ -78,9 +81,21 @@
...
+    ->set('label', $label)
+    ->set('id', $id)
+    ->set('status', TRUE)
+    ->set('description', "Test provisioner $i")
+    ->set('token', 'Later')
+    ->save();
+}
+
+$nodes = $etm->getStorage('node')
+  ->loadByProperties(['title' => 'Drupal example']);
```
behzad1q, line 101: according to line 106 this file should be 'Example'

DrupalContentGenerate.php
```diff
@@ -114,9 +129,8 @@
   echo "Project already setup.\n";
 }
 
-$nodes = \Drupal::entityTypeManager()
-    ->getStorage('node')
-    ->loadByProperties(['title' => 'Drupal test site']);
+$nodes = $etm->getStorage('node')
+  ->loadByProperties(['title' => 'Drupal test site']);
```
behzad1q, line 136: not sure if loadBy Properties is case sensitive. but better to type similar text according to line 141. 'Drupal Test Site'

dsh
```diff
@@ -119,6 +119,15 @@ dsh_openshift_init() {
     oc policy add-role-to-user admin system:serviceaccount:shepherd:shepherd
   fi
 
+  # Creating additional service accounts for namespace manipulation.
+  echo "Ensuring serviceaccounts exist for provisioning."
+  for i in {0..4} ; do
```
behzad1q, line 134: the number 4 should be read from config as it is used in DrupalContentGenerate.php:85 too

web/modules/custom/shepherd/shp_orchestration/src/OpenShiftClientFactory.php
behzad1q: how this can happen? for example I see OpenShiftOrchestrationProvider->createdEnvironment at https://github.com/universityofadelaide/shepherd/blob/b54ccaa434ced1a51b8dbd2d1e7bd624216b0a84/web/modules/custom/shepherd/shp_orchestration/src/Plugin/OrchestrationProvider/OpenShiftOrchestrationProvider.php#L240 has the $this->client and does all jobs in that function and create an environment. if you want OpenShiftOrchestrationProvider->updatedEnvironment also use the same $this->client so you need to use cache or static variable.

behzad1q:
it's good to update WordpressContentGenerate.php too

dsh
pingers: This is WIP. Per Simon's comment, it's being completely refactored... Let's not worry too much about the review for now ;)

web/modules/custom/shepherd/shp_orchestration/src/TokenNamespaceTrait.php
```diff
@@ -0,0 +1,44 @@
...
+    return base64_decode($secret['data']['token']);
+  }
+
+  /**
+   * Set the site project/namespace.
+   *
+   * @param int $site_id
+   *   The site which dictates which service account quota will be used.
+   */
+  private function setSiteNamespace(int $site_id) {
```
pingers: Isn't this "get" not "set"?

web/modules/custom/shepherd/shp_orchestration/src/TokenNamespaceTrait.php
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * Set the site token.
+   *
+   * @param int $site_id
+   *   The site which dictates which service account quota will be used.
+   *
+   * @throws \UniversityOfAdelaide\OpenShift\ClientException
+   */
+  private function setSiteToken(int $site_id) {
```
pingers: Isn't this "get" not "set"?

web/modules/custom/shepherd/shp_orchestration/src/OpenShiftClientFactory.php
singularo: Yes, decided to not do it with a static var.

dsh
singularo: These are totally for local development, and used in the `dsh` shell script and `DrupalContentGenerate.php`, no nice/easy way to share between, I don't think we need to make it more flexible.

web/modules/custom/shepherd/shp_orchestration/src/TokenNamespaceTrait.php
singularo: Yep, changed.

web/modules/custom/shepherd/shp_orchestration/src/TokenNamespaceTrait.php
singularo: Yep, changed.

web/modules/custom/shepherd/shp_orchestration/src/OpenShiftClientFactory.php
singularo: Code no longer there.

web/modules/custom/shepherd/shp_service_accounts/src/Form/ServiceAccountForm.php
singularo: Not sure what you mean, this is generated code, and a constant defined in common.inc, will just leave as it is.

web/modules/custom/shepherd/shp_service_accounts/src/Service/ServiceAccountRandomSelection.php
singularo: Made custom exceptions so they are catchable.

DrupalContentGenerate.php
singularo: Fixed this.

DrupalContentGenerate.php
singularo: Fixed this.

singularo:
Merging this to move onto the next task

PR #186, opened 2022-06-09
SAPP-2500 route promotion label

singularo:
Add labels and annotations to the generated/test code.
Add support for labels on routes to Shepherd.

### DIFF
--- a/DrupalContentGenerate.php
+++ b/DrupalContentGenerate.php
@@ -64,6 +64,18 @@ if (!$development = $etm->getStorage('taxonomy_term')->loadByProperties(['vid' =
     'vid'                   => 'shp_environment_types',
     'name'                  => 'Dev',
     'field_shp_base_domain' => $domain_name,
+    'field_shp_annotations' => [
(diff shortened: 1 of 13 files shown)


DrupalContentGenerate.php
```diff
@@ -64,6 +64,18 @@
     'vid'                   => 'shp_environment_types',
     'name'                  => 'Dev',
     'field_shp_base_domain' => $domain_name,
+    'field_shp_annotations' => [
+      [
+        'key' => 'haproxy.router.openshift.io/ip_whitelist',
```
pingers, line 69: We should leave this out - not going to be used in 4.x anyway right?

singularo: Presume we dont need it, I reckon leave it in as an example, but commented.

singularo: Presume we dont need it, I reckon leave it in as an example, but commented.

pingers:
:ship:

PR #187, opened 2022-06-10
SAPP 2796 set pvc size

singularo:
- Add storage field to site.
- Use OpenShift size suffix.
- Add storage parameters to the default site.
- Add storage size support.
- Fix render issue if an environment failed creation.
- Need to switch to site token/namespace to delete memcached pods.
- Rename field for clearer meaning.
- Simplify and make safer.
- Add backup size to the project type.
- Specify the dns, should be safe for all.
- Get the backup pvc size from the project and apply it.
- Add in backup and storage site. Keep class in but commented.
- Missed the field definitions.


pingers:
:shipit:

PR #189, opened 2022-06-20
Fixup error found in testing with terminal links.

singularo:


singularo:
The new backup operator meant we could remove some no longer used parameters as well.

pingers:
I can't fault it... :ship:

PR #198, opened 2022-08-04
Copying things from the operator across for easier dev.

singularo:
Copy across the operator rbac etc things.


singularo:
Merging this for finalising the operator debugging.

PR #200, opened 2022-08-12
Split out datagrid

singularo:
Datagrid will be in its own namespace, we need to do the changes to support that before we can deploy.
So far this splits things up ok, and `robo dev:drupal-content-generate` runs, but there are things now not working:

- [x] datagrid config not being updated, likely as it doesn't know the namespace.
- [x] memcache pod not deployed for dev container, probably because of the previous point


singularo:
@pingers This moves the datagrid out into its own project locally, and means we can do same on deployed version.

DrupalContentGenerate.php
```diff
@@ -43,11 +43,13 @@
...
 
+// Set orchestration provider config.
 $openshift_config = [
-  'endpoint'   => $openshift_url,
-  'token'      => $token,
-  'namespace'  => 'shepherd',
-  'verify_tls' => FALSE,
+  'endpoint'           => $openshift_url,
+  'token'              => $token,
+  'namespace'          => 'shepherd-uat',
```
singularo, line 50: Hmm a bit inconsistent here. Could switch all -uat to -local everywhere I guess. Sort of makes more sense.

DrupalContentGenerate.php
```diff
@@ -43,11 +43,13 @@
...
+  'site_deploy_prefix' => 'shp-local-',
```
singularo: Hmm a bit inconsistent here. Could switch all -uat to -local everywhere I guess. Sort of makes more sense.

singularo: Going to switch all to shepherd-dev here so as to be sure to not interfere with any deployed instances.

singularo:
Replacing places where we just had the word `shepherd` in various places with `shepherd-dev` and `shepherd-sa` gives a much better ability to search/replace specific things as well. Will merge this tomorrow morning unless I hear different.

PR #204, opened 2022-09-05
Update deployment.yaml

nickschuch:


deploy/prod/deployment.yaml
```diff
@@ -55,16 +53,14 @@ spec:
             value: "/home/shepherd"
           - name: SHEPHERD_OPERATOR_RESTORE_MYSQL_IMAGE
             value: "skpr/mtk-mysql"
-          - name: SHEPHERD_OPERATOR_RESTORE_AWS_ENDPOINT
```
singularo, line 58: Should we just comment that this is only required for local/different endpoint testing?

PR #207, opened 2022-09-14
Switch from random to least used service account selection.

singularo:
Spread from random seems a bit biased, this will always use the lowest used account instead.

### DIFF
--- a/web/modules/custom/shepherd/shp_service_accounts/src/Service/ServiceAccountsInterface.php
+++ b/web/modules/custom/shepherd/shp_service_accounts/src/Service/ServiceAccountsInterface.php
@@ -60,31 +60,6 @@ interface ServiceAccountsInterface {
    * @return \Drupal\shp_service_accounts\Entity\ServiceAccount|null
    *   The ServiceAccount or NULL
    */
-  public function getRandomServiceAccount(): ?ServiceAccount;
-
-  /**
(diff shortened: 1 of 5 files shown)


singularo:
We probably want to get this in early next week @pingers as we're already a little out of balance with the random one with (adjusted to remove uat):
00n - 5
01n - 5
02n - 9
So seems not truly random - surprise ;-)

web/modules/custom/shepherd/shp_service_accounts/src/Service/ServiceAccountsInterface.php
```diff
@@ -48,6 +62,17 @@ public function getServiceAccount($siteId): ?ServiceAccount;
...
+  /**
+   * Retrieve the list of service accounts and return the least used one.
+   *
+   * @throws \Exception
+   *   If there are no service accounts defined.
+   *
+   * @return \Drupal\shp_service_accounts\Entity\ServiceAccount|null
+   *   The ServiceAccount or NULL
+   */
+  public function getLowestServiceAccount(): ?ServiceAccount;
```
pingers: These don't belong in the interface.. should just be the generic getServiceAccount() and have 2 classes with the different implementations.

pingers: I'm happy to take a stab at this one.

PR #211, opened 2022-11-02
Create and use memcached imagestream from shepherd namespace.

singularo:
* Change default backend for the content generate to datagrid yml.
* Add creation of memcached imagestream to dsh, but commented.
* Add code in the memcache plugin to create and/or use memcache imagestream from shepherd namespace.

### DIFF
--- a/web/modules/custom/shepherd/shp_cache_backend/src/Plugin/CacheBackend/MemcachedDatagridYml.php
+++ b/web/modules/custom/shepherd/shp_cache_backend/src/Plugin/CacheBackend/MemcachedDatagridYml.php
@@ -440,6 +440,10 @@ class MemcachedDatagridYml extends CacheBackendBase {
    *   Memcache name.
    * @param string $memcached_port
    *   Memcache port.
+   * @param string $image_stream_tag
+   *   The image stream tag to use.
+   * @param string $image_stream_namespace
+   *   The namespace to load the image stream from.
    * @param array $data
    *   Array of data for labels.
    *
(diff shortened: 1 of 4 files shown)


web/modules/custom/shepherd/shp_cache_backend/src/Plugin/CacheBackend/MemcachedDatagridYml.php
```diff
@@ -515,7 +534,8 @@ protected function generateDeploymentConfig(NodeInterface $environment, string $
               ],
               'from' => [
                 'kind' => 'ImageStreamTag',
-                'name' => 'memcached:alpine',
+                'name' => $image_stream_tag,
+                'namespace' => $image_stream_namespace,
```
pingers, line 535: Can we patch this part of our existing memcache deployments and delete the orphaned imagestreams from site projects?

singularo: Probably, we can just deploy a single new env and it will create the memcache instance and give the deploymentconfig that we can then steal bits from.

web/modules/custom/shepherd/shp_cache_backend/src/Plugin/CacheBackend/MemcachedDatagridYml.php
```diff
@@ -335,19 +335,34 @@ public function onEnvironmentCreate(NodeInterface $environment) {
...
+    $siteToken = $this->getSiteToken($environment->field_shp_site->entity->id());
+    $siteNamespace = $this->buildProjectName($environment->field_shp_site->entity->id());
+    $this->client->setToken($siteToken);
+    $this->client->setNamespace($siteNamespace);
 
     $memcachedDeploymentName = self::getMemcachedDeploymentName($environment);
     $deploymentName = OpenShiftOrchestrationProvider::generateDeploymentName($environment->id());
     $memcachedPort = 11211;
 
+    // Image streams are in the shepherd project.
```
pingers, line 338: Maybe refactor this to another method and call it from the top of this one? Just thinking about single responsibilities.

pingers: provisionMemcacheImagestream() - something like that.

pingers: Moving to top to avoid switching back and forward between projects

PR #212, opened 2022-11-09
Disable old cache backends.

singularo:
* Provide status field so that cache backends can be enabled/disabled.
* Filter the returned available backends based on the status.
* Apply patch that removes the weird handling when a single instance.
* Apply status field to the existing cache backends.


pingers:
:ship:

PR #220, opened 2025-02-10
Feature/prevent concurrent db actions

audave:


web/modules/custom/shepherd/shp_time_restrictions/config/install/block.block.timerestrictionnotification.yml
pingers, line 1: Doesn't conform to the naming practices for shepherd blocks - block.block.shp_time_restriction_notification.yml

This block isn't weird or special either afaik in terms of dependencies, so I think we'd want to track it in config-export rather than config-install folder.

web/modules/custom/shepherd/shp_time_restrictions/shp_time_restrictions.install
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * @file
+ * Install, update, and uninstall functions for the shp_time_restrictions module.
+ */
+
+/**
+ * Implements hook_schema().
+ */
+function shp_time_restrictions_schema() {
+  $schema['shp_time_restrictions_log'] = [
```
pingers, line 12: Hmmm, this looks waaay too complicated to me.

Couldn't we just have stored a timestamp when the last action happened in config? I get we lose the context and history.. but it's just not that important - also, you derive all that context from information stored on the nodes anyway.

pingers: Actually state API is way more suited to this task. https://www.drupal.org/docs/develop/drupal-apis/state-api/state-api-overview

web/modules/custom/shepherd/shp_time_restrictions/shp_time_restrictions.module
```diff
@@ -0,0 +1,69 @@
...
+ * Logs when an "environment" node is deleted.
+ */
+function shp_time_restrictions_node_delete(NodeInterface $node) {
+  if ($node->bundle() === 'shp_environment') {
+    \Drupal::service('shp_time_restrictions.actions_log')->logAction($node, 'deleted');
+  }
+}
+
+/**
+ * Implements hook_node_delete().
```
pingers, line 23: Wrong hook copy/pasta.

web/modules/custom/shepherd/shp_time_restrictions/shp_time_restrictions.install
audave: The last action is the node creation / deletion.

You can get the last created date from the node table, but if an environment is deleted that timestamp isn't stored anywhere.

web/modules/custom/shepherd/shp_time_restrictions/src/TimeRestrictionFormAlters.php
```diff
@@ -0,0 +1,78 @@
...
+   */
+  public function preventConcurrentActionsFormAlter(array &$form, FormStateInterface $form_state, int $nid = NULL) {
+    $form['#validate'][] = [
+      TimeRestrictionFormAlters::class,
+      'preventConcurrentDeploymentsValidate',
+    ];
+
+    if (!$this->actionsLogService->isOutsideEnvironmentCreationTimeDelay($nid)) {
+      \Drupal::messenger()->addError(
+        t('To prevent errors it is not possible to create an environment at this time')
```
pingers, line 62: Language only deals with create, but delete is also possible ("create or delete" would be fine)

web/modules/custom/shepherd/shp_time_restrictions/src/TimeRestrictionFormAlters.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  /**
+   * Validate form can not be submitted after a recent environment creation.
+   */
+  public static function preventConcurrentDeploymentsValidate(array &$form, FormStateInterface $form_state) {
+    if (!\Drupal::service('shp_time_restrictions.actions_log')->isOutsideEnvironmentCreationTimeDelay()) {
+      \Drupal::messenger()->addError(t('To prevent errors it is not possible to create an environment at this time'));
+      $form_state->setErrorByName('', 'To prevent errors it is not possible to create an environment at this time');
```
pingers, line 74: Can we say when it will be okay to make changes?

pingers: Or is that already happening in the block and this is just the "STAAHP, you're doing it wrong" message?

